PR #5, opened 2019-03-26
Make local-storage work with openshift-4.0

gnufied:
This includes many fixes:

1. rename `LocalStorageProvider` to `LocalVolume` because that name ran into k8s resource name limitations of 64 chars when combined with group, name and kind.
2.  Make various fixes to YAMLs for OLM. Since this operator needs to watch for cluster level objects, it needs different Install Plan.
3. Use deviceID whenever available.


pkg/apis/local/v1alpha1/types.go
```diff
@@ -6,25 +6,30 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	defaultDiskMakerImageVersion = "quay.io/gnufied/local-diskmaker:latest"
+	defaultProvisionImage        = "quay.io/gnufied/local-volume-provisioner:v2.0.1-hemant"
```
gnufied: I have to change location of these images. I can do in a follow up.

pkg/diskmaker/diskmaker.go
```diff
@@ -125,27 +131,85 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
+	if err != nil {
+		return "", "", fmt.Errorf("error finding disk %s with %v", deviceID, err)
+	}
+	for _, disk := range allDisks {
+		drive, err := filepath.EvalSymlinks(disk)
+		if err != nil {
+			continue
+		}
+		diskNameParts := strings.Split(drive, "/")
+		realDiskName := strings.TrimSpace(diskNameParts[len(diskNameParts)-1])
```
jsafrane: `filepath.Base()`?

pkg/diskmaker/diskmaker.go
```diff
@@ -125,27 +131,85 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
+
+		diskIDParts := strings.Split(disk, "/")
+		realDiskID := strings.TrimSpace(diskIDParts[len(diskIDParts)-1])
```
jsafrane: filepath.Base()

pkg/diskmaker/diskmaker.go
```diff
@@ -125,27 +131,85 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
 }
 
+// findDeviceByID finds device ID and return name and path
+func (d *DiskMaker) findDeviceByID(deviceID string) (string, string, error) {
+	allDisks, err := filepath.Glob(diskByIDPath)
+	if err != nil {
+		return "", "", fmt.Errorf("error finding disk %s with %v", deviceID, err)
+	}
+	for _, disk := range allDisks {
+		drive, err := filepath.EvalSymlinks(disk)
```
jsafrane: It's quite confusing to mix "disk" and "drive" - they mean basically the same.

pkg/diskmaker/diskmaker.go
```diff
@@ -125,27 +131,85 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
+				continue
 			}
+			addDiskToMap(storageClass, diskPath, deviceName)
 		}
 	}
 	return blockDeviceMap, nil
 }
 
+// findDeviceByID finds device ID and return name and path
+func (d *DiskMaker) findDeviceByID(deviceID string) (string, string, error) {
```
jsafrane: If I understand it correctly (and I am not sure because the code is confusing), this function gets stable device (XXX from `/dev/disk/by-id/XXX`) and finds its corresponding `/dev/YYY`. Why can't you just do `EvalSymlinks(XXX)` and scan whole /dev/disk/by-id instead?

pkg/diskmaker/diskmaker.go
```diff
@@ -125,27 +131,85 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
+	if err != nil {
+		return "", fmt.Errorf("error finding disk %s with %v", diskName, err)
+	}
+	for _, disk := range allDisks {
+		drive, err := filepath.EvalSymlinks(disk)
+		if err != nil {
+			continue
+		}
+		diskNameParts := strings.Split(drive, "/")
+		realDiskName := strings.TrimSpace(diskNameParts[len(diskNameParts)-1])
```
jsafrane: filepath.Base()

pkg/diskmaker/diskmaker.go
gnufied: yeah that is simpler. I will change it.

pkg/diskmaker/diskmaker.go
gnufied: fixed

PR #17, opened 2019-06-20
Various changes for GA

gnufied:
This PR contains following changes:

1. Remove alpha API
2. Conform to operator API defined in `openshift/api/operator/v1/types.go`.
3. Honour managment state and record failing conditions etc.
4. Read local-storage-static-provisioner and diskmaker-image from ENV variable defined in CSV so as these can be replaced when the operator is moved to OCP/aos-art builds.


pkg/apis/local/v1/types.go
```diff
@@ -95,6 +95,13 @@ type LocalVolumeStatus struct {
 
 	// Conditions is a list of conditions and their status.
 	Conditions []operatorv1.OperatorCondition
+
+	// readyReplicas indicates how many replicas are ready and at the desired state
+	ReadyReplicas int32 `json:"readyReplicas"`
```
jsafrane: This probably belongs to commit "Add new fields to v1 api type"

pkg/controller/controller.go
```diff
@@ -162,14 +162,40 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
...
+		LastTransitionTime: metav1.Now(),
+		Message:            message,
+	}
+	if !success {
+		operatorCondition.Type = operatorv1.OperatorStatusTypeFailing
+		operatorCondition.Status = operatorv1.ConditionTrue
+		operatorCondition.LastTransitionTime = metav1.Now()
+		operatorCondition.Message = message
+	}
+	lv.Status.Conditions = append(lv.Status.Conditions, operatorCondition)
```
jsafrane: This will keep adding new conditions instead of replacing the old one(s)

pkg/controller/controller.go
```diff
@@ -162,14 +162,40 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
...
+
+func (h *Handler) addCondition(lv *localv1.LocalVolume, success bool, message string) {
+	operatorCondition := operatorv1.OperatorCondition{
+		Type:               operatorv1.OperatorStatusTypeAvailable,
+		Status:             operatorv1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
+		Message:            message,
+	}
+	if !success {
+		operatorCondition.Type = operatorv1.OperatorStatusTypeFailing
```
jsafrane: Why there are two conditions (Available and Failing) instead of single Available condition that gets `true` or `false`?

pkg/controller/controller.go
```diff
@@ -162,14 +162,40 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
...
+	if syncErr != nil {
+		logrus.Errorf("error syncing condition : %v", syncErr)
+	}
+	return err
+}
+
+func (h *Handler) addCondition(lv *localv1.LocalVolume, success bool, message string) {
+	operatorCondition := operatorv1.OperatorCondition{
+		Type:               operatorv1.OperatorStatusTypeAvailable,
+		Status:             operatorv1.ConditionTrue,
```
jsafrane: Should this condition be true even on error?

pkg/apis/local/v1/types.go
```diff
@@ -1,14 +1,19 @@
...
+
 	operatorv1 "github.com/openshift/api/operator/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	defaultDiskMakerImageVersion = "registry.svc.ci.openshift.org/ocp/4.2:local-storage-diskmaker"
-	defaultProvisionImage        = "registry.svc.ci.openshift.org/ocp/4.2:sig-storage-local-static-provisioner"
+	defaultDiskMakerImageVersion = "registry.svc.ci.openshift.org/origin/4.2:local-storage-diskmaker"
```
jsafrane: IMO,  quay.io/openshift/origin-local-storage-diskmaker looks nicer as default. In theory, they should be the same though.

And I am not sure that baking version (4.2) in code is a good idea, we probably don't want to bump it before every release.

pkg/apis/local/v1alpha1/types.go
```diff
@@ -19,6 +19,38 @@ type LocalVolumeList struct {
 	Items           []LocalVolume `json:"items"`
 }
 
+type ManagementState string
```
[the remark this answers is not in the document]
gnufied: I think this was a mistake. I did not intent to change alpha API. I think I am going to delete the whole folder.

pkg/apis/local/v1/types.go
gnufied: Yeah using quay.io/openshift image works. I am just trying to figure out if we can use quay.io/openshift image for sig-storage-local-static-provisioner too otherwise we might have to use upstream image..

pkg/apis/local/v1alpha1/types.go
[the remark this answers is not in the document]
gnufied: fixed

pkg/controller/controller.go
gnufied: fixed

pkg/controller/controller.go
gnufied: It can go either way. I think using two different conditions is easier for readability.

pkg/controller/controller.go
gnufied: I have redid this

gnufied:
cc @j-griffith

deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: local-storage-operator
       containers:
         - name: local-storage-operator
-          image: quay.io/gnufied/local-storage-operator:v0.0.8
+          image: quay.io/gnufied/local-storage-operator:v0.0.13
```
j-griffith: still no "official" image?

pkg/apis/local/v1/types.go
```diff
@@ -82,24 +95,44 @@ type LocalVolumeStatus struct {
...
+
+	// readyReplicas indicates how many replicas are ready and at the desired state
+	ReadyReplicas int32 `json:"readyReplicas"`
+
+	// generations are used to determine when an item needs to be reconciled or has changed in a way that needs a reaction.
+	// +optional
+	Generations []operatorv1.GenerationStatus `json:"generations,omitempty"`
 }
 
 // SetDefaults sets image defaults
```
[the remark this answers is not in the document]
gnufied: I am going to open a follow up PR and remove image fields from `LocalVolume` CR. I have been thinking having these fields on `LocalVolume` CR is problematic. For example -  what happens when OLM provides new manifest for the operator and these CRs are hardcoded to use a different version?

pkg/controller/api_updater.go
```diff
@@ -0,0 +1,73 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ApiUpdater interface {
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/controller/api_updater.go
```diff
@@ -0,0 +1,73 @@
...
+type SdkApiUpdater struct{}
+
+func (s *SdkApiUpdater) syncStatus(oldInstance, newInstance *localv1.LocalVolume) error {
+	logrus.Info("Syncing LocalVolume.Status")
+
+	if !equality.Semantic.DeepEqual(oldInstance.Status, newInstance.Status) {
+		logrus.Info("Updating LocalVolume.Status")
+		err := sdk.Update(newInstance)
+		if err != nil && errors.IsConflict(err) {
+			err = nil
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/controller/controller.go
```diff
@@ -93,40 +93,50 @@ func (h *Handler) syncLocalVolumeProvider(instance *v1alpha1.LocalVolume) error
...
+		logrus.Infof("operator is not managing local volumes : %v", o.Spec.ManagementState)
+		o.Status.State = o.Spec.ManagementState
+		err = h.apiUpdater.syncStatus(instance, o)
+		if err != nil {
+			return fmt.Errorf("error syncing status %v", err)
+		}
+		return nil
+	}
+
 	err = h.syncRbacPolicies(o)
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/controller/controller.go
```diff
@@ -138,73 +148,101 @@ func (h *Handler) syncLocalVolumeProvider(instance *v1alpha1.LocalVolume) error
...
 
-	if !equality.Semantic.DeepEqual(oldInstance.Status, newInstance.Status) {
-		logrus.Info("Updating LocalVolume.Status")
-		err := sdk.Update(newInstance)
-		if err != nil && errors.IsConflict(err) {
-			err = nil
+	for _, condition := range oldConditions {
+		if success {
+			// for successful condition we preserve all old conditions except successful one
+			if condition.Type != operatorv1.OperatorStatusTypeAvailable {
```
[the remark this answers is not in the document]
gnufied: okay I did this on purpose but I have fixed it up.

pkg/controller/controller.go
```diff
@@ -138,73 +148,101 @@ func (h *Handler) syncLocalVolumeProvider(instance *v1alpha1.LocalVolume) error
...
+				newConditions = append(newConditions, condition)
+			}
+		} else {
+			// for failing condition we preserve all old conditions except failed one
+			if condition.Type != operatorv1.OperatorStatusTypeFailing {
```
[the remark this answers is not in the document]
gnufied: fixed

deploy/operator.yaml
gnufied: Updated these to more official OKD images.

gnufied:
@jsafrane addresses the open comments PTAL.

j-griffith:
looks good to me assuming it covers the items pointed out by @jsafrane

PR #18, opened 2019-06-27
Remove fields from v1 api

gnufied:
This PR uses https://github.com/openshift/local-storage-operator/pull/17 as a base and:

1. Removes configurable local-storage-provisioner and diskmaker images from CR.
2. Removes separate deviceIDs and deviceName field from CR.


jsafrane:
lgtm-ish

PR #19, opened 2019-06-27
Productionize local-storage-operator

gnufied:


jsafrane:
/lgtm
I wonder how this works :-)

PR #21, opened 2019-06-28
fix a bug with symlink creation

gnufied:
Add hacking docs and update existing docs

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,44 @@
+# Notes for hacking on local-storage-operator
+
+## For developing on plain k8s
+
+Developing on plain k8s is pretty simple. Just write your code and use Makefile to build your own images
+and then update `deploy/operator.yaml` to point to your images and follow rest of the instructions in `deploy/README.md`.
+
+## For developing on Openshift and OLM
+
+1. Same as plain k8s, make your code change and build local-diskmaker and local-storage-operator images. Push them to quay or docker.io.
+2. Next we will have to update CSV inside manifests directory to point to those images.
+3. After updating the manifests file, you need to build your own local-registry. You can use `Dockerfile.registry` to do that.
+
+```
+docker build --no-cache -t quay.io/gnufied/local-registry:latest -f ./Dockerfile.registry
+```
+
+Push the result image somewhere.
+
+4. When creating a catalog from `examples/olm/catalog-create-subscribe.yaml` file, specify your own image of local registry.
+
+5. Proceed with creating CR and start using the operator.
+
+### Cleaning up after a deploy
+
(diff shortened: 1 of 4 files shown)


pkg/diskmaker/diskmaker.go
```diff
@@ -132,7 +132,12 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
 				logrus.Errorf("error creating symlink directory %s with %v", symLinkDirPath, err)
 				continue
 			}
-			symLinkPath := path.Join(symLinkDirPath, deviceNameLoction.diskNamePath)
+			baseDeviceName := filepath.Base(deviceNameLoction.diskNamePath)
```
j-griffith: Looks like these are misspelled everywhere so won't hurt, but might want to s/deviceNameLoction/deviceNameLocation/g

j-griffith: do you want to sanity check for `.` in the return from filepath.Base incase nothing is provided?  Or is that already covered and I didn't see it?

HACKING.md
```diff
@@ -0,0 +1,44 @@
...
+and then update `deploy/operator.yaml` to point to your images and follow rest of the instructions in `deploy/README.md`.
+
+## For developing on Openshift and OLM
+
+1. Same as plain k8s, make your code change and build local-diskmaker and local-storage-operator images. Push them to quay or docker.io.
+2. Next we will have to update CSV inside manifests directory to point to those images.
+3. After updating the manifests file, you need to build your own local-registry. You can use `Dockerfile.registry` to do that.
+
+```
+docker build --no-cache -t quay.io/gnufied/local-registry:latest -f ./Dockerfile.registry
```
dmoessne: can't this been done with buildah/podman ?

examples/olm/create-cr.yaml
```diff
@@ -15,4 +15,4 @@ spec:
       volumeMode: Filesystem
       fsType: xfs
       deviceNames:
-        - xvdf
+        - /dev/xvdf
```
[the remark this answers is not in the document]
gnufied: yeah we are. This was decided in conversation with CNV and OCS teams and internally openshift-storage team also kinda agreed that, this might not be a bad idea.

HACKING.md
gnufied: I am using podman.. and these instructions work identical for podman too.

examples/olm/create-cr.yaml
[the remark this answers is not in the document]
jsafrane: Is the change reflected in docs? Are there any docs? :-)

gnufied:
going to udpate other related docs.

/hold

PR #25, opened 2019-06-28
Fix makefile docker build and add push

j-griffith:
The docker build was missing the required `.`, this patch adds it and
also while we're at it we add image push functionality.  So now set 
REGISTRY env variable appropriately and do:
`make images` to build images, and `make push-images` to push the 
image to your registry.  Or, just run `make push` and do it all in one shot.


gnufied:
Yeah. I was using podman for development locally and it does not require the `.` at the end of the command, so I kinda skipped it. thanks for catching and fixing it.

PR #26, opened 2019-06-28
Update docs to use new device Path

gnufied:


gnufied:
@j-griffith Yes multiple storage classes are still supported. The generated configmaps are not user facing component and hence I updated them to be more up to date with what we serialize on disk.

PR #27, opened 2019-06-29
Update olm doc to use new devicePaths param

j-griffith:
Just switched the deploy-with-olm doc to use the new devicePaths and full path to device syntax.


gnufied:
@j-griffith is this still needed? I thought we updated this doc already.

PR #28, opened 2019-07-02
Move the operator examples to stable channel

gnufied:
- Fix bug with API version.
- Delete image-references file. It appears to be the main reason for manifests not working with operator-registry


gnufied:
Filed a bug for operator-registry - https://bugzilla.redhat.com/show_bug.cgi?id=1726409

gnufied:
There may be other way of fixing problem with operator-registry than deleting image-references file.

/hold

PR #30, opened 2019-07-03
Add validation

gnufied:
- Also rename managmentState key.
- Fix klog initialization when logrus is also present

### DIFF
--- a/manifests/4.2.0/local-volumes.crd.yaml
+++ b/manifests/4.2.0/local-volumes.crd.yaml
@@ -11,3 +11,86 @@ spec:
     singular: localvolume
   scope: Namespaced
   version: v1
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            nodeSelector:
+              description: Nodes on which the provisioner must run
+              type: object
+            storageClassDevices:
+              description: List of storage class and devices they can match
+              items:
+                properties:
+                  storageClassName:
+                    description: StorageClass name to use for set of matched devices
+                    type: string
+                  volumeMode:
+                    description: Volume mode. Block or Filesystem
+                    enum:
+                      - Block
+                      - Filesystem
+                    type: string
+                  fsType:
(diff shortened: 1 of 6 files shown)


manifests/4.2.0/local-volumes.crd.yaml
```diff
@@ -11,3 +11,86 @@ spec:
...
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
```
jsafrane: missing managementState and logLevel

manifests/4.2.0/local-volumes.crd.yaml
```diff
@@ -11,3 +11,86 @@ spec:
...
+                required:
+                  - storageClassName
+                  - devicePaths
+                type: object
+              type: array
+          required:
+            - storageClassDevices
+          type: object
+        status:
+          properties:
```
jsafrane: missing conditions and readyReplicas

gnufied:
@jsafrane fixed. PTAL

gnufied:
Also verified that the operator still works and CRs are created after this change:

```
apiVersion: local.storage.openshift.io/v1
kind: LocalVolume
metadata:
  creationTimestamp: "2019-07-03T18:43:51Z"
  generation: 69
  name: example
  namespace: default
  resourceVersion: "513"
  selfLink: /apis/local.storage.openshift.io/v1/namespaces/default/localvolumes/example
  uid: fbaf930e-09c0-4795-80ae-9737687f30ba
spec:
  logLevel: Normal
  managementState: Managed
  storageClassDevices:
  - devicePaths:
    - /dev/vde
    - /dev/vdf
    fsType: ext4
    storageClassName: foobar
    volumeMode: Filesystem
status:
  conditions:
  - lastTransitionTime: "2019-07-03T18:45:16Z"
    message: Ready
    status: "True"
    type: Available
  generations:
  - group: apps
    hash: ""
    lastGeneration: 1
    name: example-local-provisioner
    namespace: default
    resource: DaemonSet
  - group: apps
    hash: ""
    lastGeneration: 1
    name: example-local-diskmaker
    namespace: default
    resource: DaemonSet
  managementState: Managed
  observedGeneration: 68
  readyReplicas: 0

```

pkg/apis/local/v1/types.go
```diff
@@ -54,9 +54,11 @@ type StorageClassDevice struct {
 	// StorageClass name to use for set of matched devices
 	StorageClassName string `json:"storageClassName"`
 	// Volume mode. Raw or with file system
+	// + optional
 	VolumeMode PersistentVolumeMode `json:"volumeMode"`
```
jsafrane: if it is optional, should it have `omitempty`?

gnufied: fixed

gnufied:
@jsafrane can you PTAL?

gnufied:
I am gonna merge this manually, assuming all concerns were addressed. Anything outstanding we can cover in a follow up PR.

PR #33, opened 2019-07-15
Add pv ownership and diskmaker events

gnufied:
1. Adds a `local-volume-owner` label to all created PVs by local-storage-operator for easy tracking of created PVs. 
2. Report error events from node back to localvolume CR. Make sure that events aren't spammed.
3. Do not update localvolume if already on latest version and no change.
4. Allow e2e tests to run against plain k8s.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -174,20 +175,19 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
 		})
 	}
 	o.Status.Generations = children
-	o.Status.ObservedGeneration = &o.Generation
 	o.Status.State = operatorv1.Managed
-	condition := operatorv1.OperatorCondition{
-		Type:               operatorv1.OperatorStatusTypeAvailable,
-		Status:             operatorv1.ConditionTrue,
-		Message:            "Ready",
-		LastTransitionTime: metav1.Now(),
-	}
-	newConditions := []operatorv1.OperatorCondition{condition}
-	o.Status.Conditions = newConditions
-	err = h.apiClient.syncStatus(instance, o)
-	if err != nil {
-		klog.Errorf("error syncing status : %v", err)
-		return fmt.Errorf("error syncing status %v", err)
+	o = h.addSuccessCondition(o)
+
+	if !equality.Semantic.DeepEqual(instance.Status, o.Status) {
+		// A generation update can cause frivolous updates to LocalVolume object
+		// we are only going to update generation when something in LocalVolume object
+		// changes.
+		o.Status.ObservedGeneration = &o.Generation
+		err = h.apiClient.syncStatus(instance, o)
+		if err != nil {
+			klog.Errorf("error syncing status : %v", err)
+			return fmt.Errorf("error syncing status %v", err)
+		}
 	}
 	return nil
 }
@@ -207,6 +207,27 @@ func (h *Handler) addFailureCondition(oldLv *localv1.LocalVolume, lv *localv1.Lo
 		klog.Errorf("error syncing condition : %v", syncErr)
 	}
 	return err
+}
+
+func (h *Handler) addSuccessCondition(lv *localv1.LocalVolume) *localv1.LocalVolume {
+	condition := operatorv1.OperatorCondition{
+		Type:               operatorv1.OperatorStatusTypeAvailable,
+		Status:             operatorv1.ConditionTrue,
+		Message:            "Ready",
+		LastTransitionTime: metav1.Now(),
+	}
+	newConditions := []operatorv1.OperatorCondition{condition}
+	oldConditions := lv.Status.Conditions
+	for _, c := range oldConditions {
+		// if operator already has success condition - don't add again
+		if c.Type == operatorv1.OperatorStatusTypeAvailable &&
+			c.Status == operatorv1.ConditionTrue &&
+			c.Message == "Ready" {
+			return lv
+		}
+	}
+	lv.Status.Conditions = newConditions
+	return lv
 }
 
 func (h *Handler) localProvisionerImage() string {
@@ -476,14 +497,22 @@ func (h *Handler) removeUnExpectedStorageClasses(cr *localv1.LocalVolume, expect
 
 func (h *Handler) generateDiskMakerConfig(cr *localv1.LocalVolume) (*corev1.ConfigMap, error) {
 	h.diskMakerConfigName = cr.Name + "-diskmaker-configmap"
-	configMapData := make(diskmaker.DiskConfig)
+	configMapData := &diskmaker.DiskConfig{
+		Disks:           map[string]*diskmaker.Disks{},
+		OwnerName:       cr.Name,
+		OwnerNamespace:  cr.Namespace,
+		OwnerKind:       localv1.LocalVolumeKind,
+		OwnerUID:        string(cr.UID),
+		OwnerAPIVersion: localv1.SchemeGroupVersion.String(),
+	}
+
 	storageClassDevices := cr.Spec.StorageClassDevices
 	for _, storageClassDevice := range storageClassDevices {
 		disks := new(diskmaker.Disks)
 		if len(storageClassDevice.DevicePaths) > 0 {
 			disks.DevicePaths = storageClassDevice.DevicePaths
 		}
-		configMapData[storageClassDevice.StorageClassName] = disks
+		configMapData.Disks[storageClassDevice.StorageClassName] = disks
 	}
 
 	configMap := &corev1.ConfigMap{
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -43,10 +43,14 @@ const (
 	// Name of the component
 	componentName = "local-storage-operator"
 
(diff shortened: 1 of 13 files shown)


pkg/controller/controller.go
```diff
@@ -170,20 +175,19 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
...
-	if err != nil {
-		klog.Errorf("error syncing status : %v", err)
-		return fmt.Errorf("error syncing status %v", err)
+	o = h.addSuccessCondition(o)
+
+	if !equality.Semantic.DeepEqual(instance.Status, o.Status) {
+		// A generation update can cause frivolous updates to LocalVolume object
+		// we are only going to update generation when something in LocalVolume object
+		// changes.
+		o.Status.ObservedGeneration = &o.Generation
```
jsafrane: I am not sure about this change. When someone changes `.spec` (= `generation` is incremented), `.status` should be changed too so users can see that the operator is doing something about their change. If `observedGeneration` is stuck at an older value it shows that something is wrong with the operator.

pkg/controller/controller.go
```diff
@@ -170,20 +175,19 @@ func (h *Handler) syncLocalVolumeProvider(instance *localv1.LocalVolume) error {
...
+		err = h.apiClient.syncStatus(instance, o)
```
jsafrane: In addition, I would put the equality check into `syncStatus`, it should be done every time something wants to update the status.

pkg/controller/controller.go
```diff
@@ -339,6 +364,52 @@ func (h *Handler) syncRBACPolicies(o *localv1.LocalVolume) error {
...
+				Verbs:     []string{"get"},
+				APIGroups: []string{"local.storage.openshift.io"},
+				Resources: []string{"*"},
+			},
+		},
+	}
+	addOwner(&localVolumeRole.ObjectMeta, o)
+	_, _, err = h.apiClient.applyRole(localVolumeRole)
+	if err != nil {
+		return fmt.Errorf("error applying localvolume role %s with %v", localVolumeRole.Name, err)
```
jsafrane: nit: `error applying localvolume role %s: %v`

pkg/controller/controller.go
```diff
@@ -339,6 +364,52 @@ func (h *Handler) syncRBACPolicies(o *localv1.LocalVolume) error {
...
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     localVolumeRole.Name,
+		},
+	}
+
+	addOwner(&localVolumeRoleBinding.ObjectMeta, o)
+	_, _, err = h.apiClient.applyRoleBinding(localVolumeRoleBinding)
+	if err != nil {
+		return fmt.Errorf("error applying localvolume rolebinding %s with %v", localVolumeRoleBinding.Name, err)
```
jsafrane: nit: `error applying localvolume rolebinding %s: %v`

pkg/controller/controller.go
```diff
@@ -371,8 +442,15 @@ func (h *Handler) generateProvisionerConfigMap(cr *localv1.LocalVolume) (*corev1
 	if err != nil {
 		return nil, fmt.Errorf("error creating configmap while marshalling yaml %v", err)
 	}
+
+	pvLabelString, err := yaml.Marshal(pvLabels(cr.Name))
+	if err != nil {
+		return nil, fmt.Errorf("error generating pv labels : %v", err)
```
jsafrane: nit: extra space before ` :`:
`error generating pv labels: %v`

pkg/diskmaker/diskmaker.go
```diff
@@ -28,6 +30,9 @@ var (
 type DiskMaker struct {
 	configLocation  string
 	symlinkLocation string
+	apiClient       apiUpdater
+	localVolume     *localv1.LocalVolume
+	er              *eventReporter
```
jsafrane: nit: `er` looks very similar to `err`

pkg/diskmaker/diskmaker.go
```diff
@@ -55,7 +62,25 @@ func (d *DiskMaker) loadConfig() (DiskConfig, error) {
...
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       diskConfig.OwnerKind,
+			APIVersion: diskConfig.OwnerAPIVersion,
+		},
+	}
+	lv, err = d.apiClient.getLocalVolume(lv)
+
+	if err != nil {
+		return nil, fmt.Errorf("error fetching local volume %s/%s with %v", diskConfig.OwnerName, diskConfig.OwnerNamespace, err)
```
jsafrane: nit: `error fetching local volume %s/%s: %v`
nit: use `namespace / name`, not the other way around

pkg/diskmaker/diskmaker.go
```diff
@@ -85,19 +110,25 @@ func (d *DiskMaker) Run(stop <-chan struct{}) {
...
+		msg := fmt.Sprintf("error running lsblk %v", err)
+		e := newEvent(ErrorRunningBlockList, msg, "")
+		d.er.report(e, d.localVolume)
+		klog.Errorf(msg)
 		return
 	}
 	deviceSet, err := d.findNewDisks(out.String())
 	if err != nil {
-		klog.Errorf("error unmrashalling json %v", err)
+		msg := fmt.Sprintf("error reading blocklist %v", err)
```
jsafrane: nit: `error reading blocklist: %v`

pkg/diskmaker/diskmaker.go
```diff
@@ -109,18 +140,27 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
 	// read all available disks from /dev/disk/by-id/*
 	allDiskIds, err := filepath.Glob(diskByIDPath)
 	if err != nil {
-		klog.Errorf("error listing disks in /dev/disk/by-id : %v", err)
+		msg := fmt.Sprintf("error listing disks in /dev/disk/by-id : %v", err)
```
jsafrane: nit: extra space before `:`

pkg/diskmaker/diskmaker.go
```diff
@@ -109,18 +140,27 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
...
+		msg := fmt.Sprintf("eror finding matching disks : %v", err)
+		e := newEvent(ErrorFindingMatchingDisk, msg, "")
+		d.er.report(e, d.localVolume)
+		klog.Errorf(msg)
 		return
 	}
 
 	if len(deviceMap) == 0 {
-		klog.Errorf("unable to find any matching disks")
+		msg := fmt.Sprintf("found empty matching device list")
```
jsafrane: nit: useless `Sprintf`, you can put it directly to `newEvent` below.

pkg/diskmaker/diskmaker.go
```diff
@@ -129,7 +169,10 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
 			symLinkDirPath := path.Join(d.symlinkLocation, storageClass)
 			err := os.MkdirAll(symLinkDirPath, 0755)
 			if err != nil {
-				klog.Errorf("error creating symlink directory %s with %v", symLinkDirPath, err)
+				msg := fmt.Sprintf("error creating symlink dir %s with %v", symLinkDirPath, err)
```
jsafrane: nit: `s/with/:/`

pkg/diskmaker/diskmaker.go
```diff
@@ -148,14 +191,21 @@ func (d *DiskMaker) symLinkDisks(diskConfig DiskConfig) {
 			}
 
 			if symLinkErr != nil {
-				klog.Errorf("error creating symlink %s with %v", symLinkPath, err)
+				msg := fmt.Sprintf("error creating symlink %s with %v", symLinkPath, err)
```
jsafrane: nit: s/with/:/

pkg/diskmaker/diskmaker.go
```diff
@@ -190,7 +240,10 @@ func (d *DiskMaker) findMatchingDisks(diskConfig DiskConfig, deviceSet sets.Stri
 		for _, deviceID := range deviceIds {
 			matchedDeviceID, matchedDiskName, err := d.findDeviceByID(deviceID)
 			if err != nil {
-				klog.Errorf("unable to add disk-id %s to local disk pool %v", deviceID, err)
+				msg := fmt.Sprintf("unable to add disk-id %s to local disk pool %v", deviceID, err)
```
jsafrane: nit: add `:` before `%v (err)`

pkg/controller/controller.go
gnufied: So, `.Generation`  is a meta field which is marked as `readonly` and only api-server increments it each time an object is saved.

So what is happening is - lets say there was no other update to `LocalVolume` object and we set `Status.ObservedGeneration` to current value of `.Generation` - but act of updating `.Status.ObservedGeneration`  will cause `.Generation` to be incremented. So this results in a infinite loop where `LocalVolume` is always being updated.  Among other things - this causes unnecessary churn  in api-server and I am not sure it is necessary.

gnufied: So in a nutshell - at any point you observe `LocalVolume` - its `ObservedGeneration` will be one number behind `.Generation` .. no matter what we do.

jsafrane: > act of updating .Status.ObservedGeneration will cause .Generation to be incremented

That should not happen. `status` should be a subresource and its change then does not bump `metadata.generation`.

> Among other things - this causes unnecessary churn in api-server and I am not sure it is necessary.

It is one of ways how to detect unresponsive operator.

jsafrane: I managed to update `status` this way:
https://github.com/openshift/csi-operator/blob/master/pkg/controller/csidriverdeployment/csidriverdeployment_controller.go#L501

And CRD gest status subresource here:
https://github.com/openshift/csi-operator/blob/master/deploy/prerequisites/01_crd.yaml#L14-L15

gnufied: Yeah - I did not mark `status` as a subresource in the CRD.  I could do that.

gnufied: Since I am not using version of operator-sdk you were using.. this will be slightly more complicated I think..

PR #35, opened 2019-07-19
{WIP} Add support for shared directories

gnufied:
Allow any directory on host to be consumed as local storage


pcsherid:
@gnufied, I see this is tagged as WIP. Is this still in development?

gnufied:
@pcsherid Yeah my attention has kinda wondered off and this feature was deemed as not something customers needed right away, so this branch is sort of frozen. If you have a use case feel free to open a github issue(or open a BZ) and we will see what we can do.

PR #37, opened 2019-07-24
Fix example of local-registry

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1732745


gnufied:
This is how the catalog entry looks after:

```
~> oc get catalogsources.operators.coreos.com
NAME                      NAME                     TYPE   PUBLISHER   AGE
local-storage-manifests   Local Storage Operator   grpc   redhat      4m26s

```

PR #42, opened 2019-09-04
Fix rbac permissions when operator is installed via OperatorSource

gnufied:
When installing via operator source, the pvc and events permissions
are granted only for local-storage namespace but local storage
provisioner requires permission at cluster scope.

/sig storage

cc @jsafrane @bertinatto

### DIFF
--- a/manifests/4.2.0/local-storage-operator.v4.2.0.clusterserviceversion.yaml
+++ b/manifests/4.2.0/local-storage-operator.v4.2.0.clusterserviceversion.yaml
@@ -151,6 +151,13 @@ spec:
           - apiGroups:
             - ""
             resources:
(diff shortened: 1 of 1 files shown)


gnufied:
There seems to be a minor difference in how operator is installed when installed via CatalogSource as documented - https://github.com/openshift/local-storage-operator/blob/master/docs/deploy-with-olm.md vs when deployed via OperatorSource (https://github.com/operator-framework/operator-marketplace/blob/master/deploy/crds/operators_v1_operatorsource_crd.yaml) . Namely it looks like when operator was installed via CatalogSource the generated rbac permissions from CSV granted cluster-wide pvc/event permissions whereas when deployed via OperatorSource the permissions are only granted for selected namespace.

PR #43, opened 2019-09-06
manifests/local-storage-operator.package.yaml: standardize channel name

sosiouxme:


sosiouxme:
@bertinatto @jsafrane these changes are necessary to get this operator working in OCP 4.2

jsafrane:
@sosiouxme, please check the failed test, something needs to be updated with the new path:

```
cat: ./hack/../deploy/crd.yaml: No such file or directory
```

sosiouxme:
@jsafrane yes, i don't understand this well enough to know what is looking at that and why. is it just a script that needs to be updated to match the changed layout?

gnufied:
The deploy script uses symlink from manifests directory into deploy directory. This PR makes the symlink invalid. We just need to fix the symlink in `deploy/crd.yaml` to point to new location.

sosiouxme:
ah, i see, very simple, sorry i missed that... trying again

sosiouxme:
basic tests at least are succeeding now, can i get a lgtm/approve?

PR #46, opened 2019-09-24
Amend docs to use redhat-local-storage-src

gnufied:


examples/olm/catalog-create-subscribe.yaml
```diff
@@ -33,5 +33,5 @@ spec:
   channel: stable
   installPlanApproval: Automatic
   name: local-storage-operator
-  source: hekumar-operators
+  source: redhat-local-storage-src
```
bertinatto: Do we also want to change line https://github.com/openshift/local-storage-operator/blob/2bd3bbed154cb80e90de35f83460f80a66db9572/examples/olm/catalog-create-subscribe.yaml#L23?

examples/olm/catalog-create-subscribe.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
 apiVersion: operators.coreos.com/v1
 kind: OperatorSource
 metadata:
-  name: hekumar-operators
+  name: redhat-local-storage-src
```
jsafrane: not sure if name of non-existing company is the right one. Origin? OpenShift? okd?

gnufied: It is not company name! it is just a internal name. The following fields is what users will see in the UI:

```
displayName: "redhat stroage operators"
  publisher: "Red Hat Storage"
```

jsafrane: ack

examples/olm/catalog-create-subscribe.yaml
gnufied: it has to be this still, because the operator is published as a application to my quay.io account. We just moved from using catalog registry which was hosted under my account to using an application still hosted under my account in quay.io

PR #51, opened 2019-10-08
Added error if a disk can't be found.

huffmanca:
Previously, if multiple devices were defined, and a node contained some of them, no error would be reported. 

This includes an event when any disk cannot be found on a node. For instance:

>   Warning  ErrorFindingMatchingDisk  2m10s  local-storage-diskmaker  ip-10-0-150-160.us-east-2.compute.internal - unable to find matching disk /dev/xvdf
>   Warning  ErrorFindingMatchingDisk  2m10s  local-storage-diskmaker  ip-10-0-150-160.us-east-2.compute.internal - unable to find matching disk /dev/xvdg


jsafrane:
/approve
/assign @gnufied 
For second pair of eyes

PR #53, opened 2019-10-30
Fix typos

gnufied:
cc @huffmanca


examples/olm/catalog-create-subscribe.yaml
```diff
@@ -21,7 +21,7 @@ spec:
   type: appregistry
   endpoint: https://quay.io/cnr
   registryNamespace: hekumar
-  displayName: "redhat stroage operators"
+  displayName: "Red Hat Storage  operators"
```
huffmanca: Should there be two spaces between "Storage" and "operators" ?

gnufied: nope.

PR #59, opened 2019-12-02
Bug 1774278: [4.3] Include Tolerations in the LocalVolume CR.

huffmanca:
Allows a user to pass Tolerations into the LocalVolume CR, which will then be propagated to the diskmaker and provisioner DaemonSets.

I've ran the unit tests, and Tolerations are included as nil if not defined. If they are defined, then these are passed to the DaemonSets and scheduled appropriately. I tested these with and without Tolerations, and they appear to schedule the pods correctly.

This is for OCP 4.3+

### DIFF
--- a/docs/deploy-with-olm.md
+++ b/docs/deploy-with-olm.md
@@ -39,7 +39,7 @@ oc describe no -l node-role.kubernetes.io/worker | grep hostname
 
 Create a LocalVolume manifest named ``create-cr.yaml`` using the hostnames obtained above:
 
-### CR using volumeMode - Filesystem
+#### CR using volumeMode - Filesystem
 
 ```yaml
 apiVersion: "local.storage.openshift.io/v1"
@@ -65,7 +65,7 @@ spec:
         - /dev/xvdf
 ```
 
-### CR using volumeMode - Block
+#### CR using volumeMode - Block
 
 ```yaml
 apiVersion: "local.storage.openshift.io/v1"
@@ -108,6 +108,36 @@ the Block CR, you may run in to issues where the pod reports being unschedulable
 ``Warning  FailedScheduling  27s (x2 over 27s)  default-scheduler  0/6 nodes are available: 3 node(s) didn't find available persistent volumes to bind, 3 node(s) had taints that the pod didn't tolerate.``
 
 In this example, the disks were prevsiously used, and a filesystem was applied.  Be sure to either use fresh disks, or remove any partitions and file systems if you're setting up a Block mode CR.
+
+### Create a CR using Tolerations
+
+In addition to a node selector, you can also specify [tolerations](https://docs.openshift.com/container-platform/latest/nodes/scheduling/nodes-scheduler-taints-tolerations.html) 
+for pods created by the local-storage-operator. This allows the local-storage-operator to select
+tainted nodes for use.
+
+The following example demonstrates selecting nodes that have a taint, `localstorage`, 
+that contains a value of `local`:
+
+```yaml
+apiVersion: "local.storage.openshift.io/v1"
+kind: "LocalVolume"
+metadata:
+  name: "local-disks"
+  namespace: "local-storage"
+spec:
+  tolerations:
+    - key: localstorage
+      operator: Equal
+      value: "local"
+  storageClassDevices:
+    - storageClassName: "local-sc"
+      volumeMode: Filesystem
+      fsType: xfs
+      devicePaths:
+        - /dev/xvdf
+```
+
+The defined tolerations will be passed to the resulting DaemonSets, allowing the diskmaker and provisioner pods to be created for nodes that contain the specified taints.
 
 ### Verify your deployment
 
--- a/pkg/apis/local/v1/types.go
+++ b/pkg/apis/local/v1/types.go
@@ -37,6 +37,9 @@ type LocalVolumeSpec struct {
 	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
 	// List of storage class and devices they can match
 	StorageClassDevices []StorageClassDevice `json:"storageClassDevices,omitempty"`
+	// If specified, a list of tolerations to pass to the diskmaker and provisioner DaemonSets.
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 }
 
 // PersistentVolumeMode describes how a volume is intended to be consumed, either Block or Filesystem.
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -675,6 +675,7 @@ func (h *Handler) generateLocalProvisionerDaemonset(cr *localv1.LocalVolume) *ap
 				Spec: corev1.PodSpec{
 					Containers:         containers,
 					ServiceAccountName: provisionerServiceAccount,
+					Tolerations:        cr.Spec.Tolerations,
 					Volumes:            volumes,
 				},
 			},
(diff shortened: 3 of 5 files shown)


gnufied:
It might be possible to verify that toleration was set on generated daemonset too, via the e2e. Also, lets update some docs for this too.

huffmanca:
@gnufied ,

I updated the `deploy-with-olm.md` file to include notes on taints and tolerations. Examining the DaemonSets using the e2e test reveal that they get none of the tolerations. Examining the DaemonSets created from the Operator outside the test reveal that they expected tolerations:

```
...
 Tolerations:    dedicated=infra
                 node.kubernetes.io/disk-pressure:NoSchedule
                 node.kubernetes.io/memory-pressure:NoSchedule
...
```

I'm looking into the e2e test further.

pkg/controller/controller.go
```diff
@@ -675,6 +675,7 @@ func (h *Handler) generateLocalProvisionerDaemonset(cr *localv1.LocalVolume) *ap
 				Spec: corev1.PodSpec{
 					Containers:         containers,
 					ServiceAccountName: provisionerServiceAccount,
+					Tolerations:	 	cr.Spec.Tolerations,
```
jsafrane: Whitespaces are odd here (or bad gihub rendering?)

jsafrane: Filed #60 to get go fmt in PR tests

jsafrane:
/retitle Bug 1774278: Include Tolerations in the LocalVolume CR.

"Bug 1774278:" format enables our BZ automation.

jsafrane:
bugzilla refresh

docs/deploy-with-olm.md
```diff
@@ -109,6 +109,37 @@ the Block CR, you may run in to issues where the pod reports being unschedulable
 
 In this example, the disks were prevsiously used, and a filesystem was applied.  Be sure to either use fresh disks, or remove any partitions and file systems if you're setting up a Block mode CR.
 
+### Create a CR using Tolerations
+
+The following example assumes that some nodes in the cluster have [taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) that designate them for use with the local-storage Operator.
```
bertinatto: Should we point to OKD docs page instead?

pkg/apis/local/v1/types.go
```diff
@@ -37,6 +37,9 @@ type LocalVolumeSpec struct {
 	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
 	// List of storage class and devices they can match
 	StorageClassDevices []StorageClassDevice `json:"storageClassDevices,omitempty"`
+	// If specified, a list of tolerations to pass to the diskmaker and provisioner DaemonSets.
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty" protobuf:"bytes,22,opt,name=tolerations"`
```
bertinatto: Is `protobuf` required here? We're not using this tag in any other field.

test/e2e/localstorage_test.go
```diff
@@ -139,6 +139,135 @@ func TestLocalStorageOperator(t *testing.T) {
...
+		},
+		Spec: localv1.LocalVolumeSpec{
+			Tolerations: []v1.Toleration{
+				{
+					Key:	"localstorage",
+					Value:	"testvalue",
+					Operator: "Equal",
+				},
+			},
+			NodeSelector: &v1.NodeSelector{
```
gnufied: Do we need this node selector?

huffmanca: I tested without the NodeSelector, and it will attempt to use all of the nodes. With NodeSelector it takes only a single node (as in the other e2e test), and then will apply/remove the taint during the test.

If the Toleration isn't applied in the LocalVolume, then it will find a node, apply the taint, and be unable to create the DS/Pod as no Node matches the criteria.

pkg/apis/local/v1/types.go
huffmanca: I initially copied it from the Pod definition. Testing with it removed reveals no difference that I've noticed. The e2e test continues to pass, and if I test manually it still schedules on the tainted node.

I've removed this bit for now.

docs/deploy-with-olm.md
huffmanca: Updated to use the `latest` version of the OpenShift docs.

pkg/controller/controller.go
huffmanca: These should now be corrected.

docs/deploy-with-olm.md
```diff
@@ -109,6 +109,37 @@ the Block CR, you may run in to issues where the pod reports being unschedulable
 
 In this example, the disks were prevsiously used, and a filesystem was applied.  Be sure to either use fresh disks, or remove any partitions and file systems if you're setting up a Block mode CR.
 
+### Create a CR using Tolerations
+
+The following example assumes that some nodes in the cluster have [taints](https://docs.openshift.com/container-platform/latest/nodes/scheduling/nodes-scheduler-taints-tolerations.html) that designate them for use with the local-storage Operator.
+
+For instance, assume that some nodes have a key, `localstorage`, with a value of `true`. With the following configuration the local-storage Operator will only search for volumes on nodes that match this key, or nodes that are generally available for scheduling:
+
```
gnufied: This documentation of taints is not entirely accurate. Tolerations don't behave like a selector and this makes it sound like one.

docs/deploy-with-olm.md
```diff
@@ -109,6 +109,37 @@ the Block CR, you may run in to issues where the pod reports being unschedulable
...
+      devicePaths:
+        - /dev/xvdf
+```
+
+The defined tolerations will be passed to the resulting DaemonSets, and will determine the available nodes for the diskmaker and provisioner pods.
+
+#### NOTE about Tolerations and NodeSelector
+
+When using both a NodeSelector and Tolerations only nodes that match both sets will be used by the local-storage Operator. Be careful with the union of these sets, as it may be more restrictive than intended.
+
```
gnufied: Same as above - toleration merely say that pods created by the operator will tolerate the specified taint if node has the same taint. I do not think they are equivalent to selectors.

docs/deploy-with-olm.md
huffmanca: Updated this section to be more explicit that this will match the specified taints, along with generally available nodes.

docs/deploy-with-olm.md
huffmanca: I've removed this note entirely.

test/e2e/localstorage_test.go
```diff
@@ -139,6 +139,135 @@ func TestLocalStorageOperator(t *testing.T) {
...
+		}
+	}
+
+	// Restore the original set of node taints after the test is completed.
+	selectedNode.Spec.Taints = nodeTaints
+	selectedNode, err = waitForNodeTaintUpdate(t, f.KubeClient, selectedNode, retryInterval, timeout)
+	if err != nil {
+		t.Fatalf("error restoring original taints on node : %v", err)
+	}
+}
```
gnufied: This test as a separate function IMO adds very little to existing test suite and copies large chunk of code from previous function. I would think that we can reuse existing function to add checks for tolerations too.  It would help keeping test time shorter and test file shorter too.

The other thing is - the test should actually verify if created daemonset has user specified toleration which it does not currently.

docs/deploy-with-olm.md
gnufied: How about a simple - "In addition to node selector, you can also specify tolerations for pods created by local-storage-operator, so as if selected nodes are tainted local-storage-operator pods can still run on those nodes" ?

test/e2e/localstorage_test.go
huffmanca: The test has been updated to run through all of the elements defined at the start of the function, and it confirm that the DaemonSets' Tolerations match the LocalVolume's Tolerations.

docs/deploy-with-olm.md
huffmanca: This text has been updated to largely incorporate your statement. I still have two paragraphs, the second leading into the example itself.

PR #61, opened 2019-12-04
Bug 1770183: Allow operator to be upgraded automatically

gnufied:
Allow operator to be updated automatically across minor versions.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1770183


gnufied:
More information about how this is supposed to work - https://docs.openshift.com/container-platform/4.2/operators/understanding-olm/olm-understanding-olm.html#olm-upgrades-replacing-multiple_olm-understanding-olm

jsafrane:
/retitle Bug 1770183: Allow operator to be upgraded automatically

PR #62, opened 2019-12-04
Bug 1779545: [4.3] Allow operator to be upgraded automatically when an update is available

jsafrane:
This is cherry pick of #61 into 4.3 branch.

/assign @gnufied


jsafrane:
Fixed PR base to 4.3
/bugzilla refresh

PR #68, opened 2019-12-09
Updating manifests to include Toleration definitions

huffmanca:
The LocalVolume CRDs were not updated as part of https://github.com/openshift/local-storage-operator/pull/59 . This PR corrects that oversight.

### DIFF
--- a/manifests/4.3/local-volumes.crd.yaml
+++ b/manifests/4.3/local-volumes.crd.yaml
@@ -68,6 +68,11 @@ spec:
                   - devicePaths
                 type: object
               type: array
+            tolerations:
+              description: A list of tolerations to pass to the diskmaker and provisioner DaemonSets.
+              items:
+                type: object
+              type: array
           required:
             - storageClassDevices
           type: object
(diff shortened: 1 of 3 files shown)


manifests/4.3/local-volumes.crd.yaml
```diff
@@ -68,6 +68,45 @@ spec:
                   - devicePaths
                 type: object
               type: array
+            tolerations:
```
gnufied: We should define how we defined node selector here. The intent is to not redefine what definition already exists in Kubernetes.

huffmanca: Understood. This has been updated to not replicate the K8s content in the list of defined items.

manifests/4.3/local-volumes.crd.yaml
```diff
@@ -68,6 +68,9 @@ spec:
...
+              description: A list of tolerations to allow local volumes to be scheduled on tainted nodes
+              type: array
```
gnufied: shouldn't this be:

```yaml
tolerations:
  descriptions: "xxxxxx"
  items:
   type: object
  type: array
```

gnufied:
We need to cherry-pick this to 4.3?

PR #70, opened 2019-12-10
4.2: Bug 1781871: Fix code for finding new disks

gnufied:
Fix finding of localvolume disks for lvm disks.


huffmanca:
I examined this to compare it to https://github.com/openshift/local-storage-operator/pull/67 (which is already merged), and it looks good to me.

jsafrane:
Set BZ labels manually, #1781871 is a valid 4.2.z bug.

jsafrane:
Wait, the bot is right... We need the 4.3 and 4.4 bugs tested first.
/bugzilla refresh

PR #73, opened 2019-12-11
[4.3] Bug 1782574: Restrict minimum kube version

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1782574

### DIFF
--- a/manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml
+++ b/manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml
@@ -48,6 +48,7 @@ spec:
     - name: Source Repository
       url: https://github.com/openshift/local-storage-operator
   version: 4.2.0
+  minKubeVersion: 1.14.0
   maturity: stable
   maintainers:
     - email: aos-storage-staff@redhat.com
--- a/manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
+++ b/manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
@@ -49,6 +49,7 @@ spec:
     - name: Source Repository
       url: https://github.com/openshift/local-storage-operator
   version: 4.3.0
+  minKubeVersion: 1.14.0
   maturity: stable
   maintainers:
     - email: aos-storage-staff@redhat.com
--- a/manifests/4.4/local-storage-operator.v4.4.0.clusterserviceversion.yaml
+++ b/manifests/4.4/local-storage-operator.v4.4.0.clusterserviceversion.yaml
@@ -49,6 +49,7 @@ spec:
     - name: Source Repository
       url: https://github.com/openshift/local-storage-operator
   version: 4.4.0
+  minKubeVersion: 1.14.0
   maturity: stable
   maintainers:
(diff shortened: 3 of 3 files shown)


manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
```diff
@@ -49,6 +49,7 @@ spec:
     - name: Source Repository
       url: https://github.com/openshift/local-storage-operator
   version: 4.3.0
+  minKubeVersion: 1.14.0
```
bertinatto: I'm wondering if we can use 1.16 for 4.3 and 1.17 for 4.4 in order to prevent people from using the wrong manifests.

bertinatto:
/hold

Adding hold in case https://github.com/openshift/local-storage-operator/pull/73#discussion_r357030984 should be addressed.

manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
gnufied: As I understand it - `minKubeVersion` is only designed to have a last minute switch to stop an operator from being installed on a incompatible kube release. I for example when tested this - this indeed prevented operator from being installed but there were no hints in the UI that operator was prevented from being installed because of kube version mis-match.  

In a nutshell - this is not a good mechanism to specify version ranges of kubernetes(or openshift). The manifests are still being displayed to the user because they come from same operatorsource.  I would rather OLM team provided a better mechanism for this. May be index images will solve it.

jsafrane: Should be 4.2 requirement noted somewhere in the operator description then?

gnufied: yeah I can update description of 4.2 operator.

jsafrane: OLM is working on a better mechanism how to restrict operator version to OCP version, https://coreos.slack.com/archives/C3VS0LV41/p1576081661098700?thread_ts=1576079878.090000&cid=C3VS0LV41

I like it as it is in this PR, let's use `minKubeVersion` to prevent users installing it on 4.1, but then it's up to the user to subscribe to the right channel. Users can install newest 4.3 operator in 4.2, IMO it should work. We must be careful when triaging customer issues, so the operator version is in sync with OCP version and suggest upgrade / downgrade otherwise.

jsafrane:
out of curiosity...
/bugzilla refresh

manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml
```diff
@@ -48,6 +48,7 @@ spec:
     - name: Source Repository
       url: https://github.com/openshift/local-storage-operator
   version: 4.2.0
+  minKubeVersion: 1.14.0
```
jsafrane: in master you removed whole `manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml`. Should we keep it in 4.3?

PR #75, opened 2019-12-16
Bug 1783835: Fix managment state displayName

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1783835


gnufied:
New display
actual_state

PR #79, opened 2020-01-07
Bug 1784316: 4.3: Fix install-mode for local-storage operator

jsafrane:
Backported https://github.com/openshift/local-storage-operator/pull/78 into 4.3


gnufied:
I have been thinking about this and I wonder if narrowing InstallMode of a operator is a supported thing during cluster upgrade. For example - if customer had 2 local-storage-operators running in 2 random namespaces and then he/she upgrades the cluster, will it break? 

/hold

jsafrane:
/close
This could break the operator in 4.3.z update.

PR #80, opened 2020-01-07
Bug 1784321: 4.2: Fix install-mode for local-storage operator

jsafrane:
Backported #78 into 4.2


gnufied:
We are not planning to fix this in 4.2 and hence closing.

PR #82, opened 2020-01-21
DiskMaker: Adds support for using all available device paths in the nodes

ashishranjan738:
This commit modifies API scheme of LocalVolumeCRD for supporting `useAllDevices` feature and also adds generated code for the above change. UseAllDevices feature will enable DiskMaker to automatically detect the available device paths in the nodes for provisioning PVs.

Signed-off-by: Ashish Ranjan <ashishranjan738@gmail.com>

### DIFF
--- a/manifests/4.4/local-volumes.crd.yaml
+++ b/manifests/4.4/local-volumes.crd.yaml
@@ -57,6 +57,21 @@ spec:
                   fsType:
                     description: File system type
                     type: string
+                  useAllDevices:
+                      description: Switch to enable selection of all available paths for storage
+                      type: boolean
+                  excludeDeviceNames:
+                    description: 'A list of regex patterns that match devices which would be excluded for local storage when useAllDevices is enabled.
+                    For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
+                    items:
+                      type: string
+                    type: array
+                  excludeDeviceTypes:
+                    description: 'A list of devices types which would be excluded for local storage when useAllDevices is enabled.
+                    For example - ["part", "lvm", "crypt"]'
+                    items:
+                      type: string
+                    type: array
                   devicePaths:
                     description: 'A list of devices which would be chosen for local storage.
                     For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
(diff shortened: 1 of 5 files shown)


gnufied:
As discussed in bjn call, we should not define options related to autodiscovery as top level fields since they can conflict with each other and in general is unnecessary.

I have been thinking about this some more. So how about - https://gist.github.com/gnufied/059fefc84c0750072ff4977cf5c4c84d ?

rohantmp:
@gnufied I think that an option like use-only-SSDs and don't-use-partitions (the feature we really need) can be combined usefully why prevent that? 

how about we have a something like

```
propertyPolicy:
  ssd: include/exclude/default
  partition: include/exclude/default

**Comment by gnufied** [comment]
@rohantmp can you define your type fully in a gist or inline.  we don't need a separate field for paritions because `DeviceTypes` can be already configured to include `[disk, part]` etc.

**Comment by leseb** [comment]
> @ashishranjan738 told me about the concern that this might leave us unable to differentiate between certain PV types. Eg. rook doesn't currently consume partitions correctly.

Yes it does in 1.2.2 but there is a bug in c-v preventing us from using partitions, it'll be fixed in Ceph 14.2.7.

File: manifests/4.4/local-volumes.crd.yaml [hunk 1]
```diff
@@ -57,6 +57,21 @@ spec:
                   fsType:
                     description: File system type
                     type: string
+                  useAllDevices:
+                      description: Switch to enable selection of all available paths for storage
+                      type: boolean
+                  excludeDeviceNames:
+                    description: 'A list of regex patterns that match devices which would be excluded for local storage when useAllDevices is enabled.
+                    For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
```
leseb: We could reject partitions too.

leseb: As well as LVs.

rohantmp: @leseb Based on a discussion yesterday, we are abandoning the idea of making this API change for now and exploring the possibility of another CRD for discovery in the future.

The reasoning was that changing the existing API with a new, more risky feature could be bad UX.

ashishranjan738:
closing this PR as we not going to change the existing APIs

PR #84, opened 2020-01-31
Bug 1793750: Fix ds force rollout

gnufied:
Ensure that daemonsets are rolled out when something in LocalVolume object changes.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1793750

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -210,11 +210,11 @@ func TestSyncLocalVolumeProvider(t *testing.T) {
 	var localProvisionerDaemonSet *appsv1.DaemonSet
 	var diskMakerDaemonset *appsv1.DaemonSet
 	for _, ds := range provisionedDaemonSets {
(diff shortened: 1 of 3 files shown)


pkg/controller/controller_test.go
```diff
@@ -73,7 +78,8 @@ func (s *fakeApiUpdater) applyStorageClass(sc *storagev1.StorageClass) (*storage
 }
 
 func (s *fakeApiUpdater) applyDaemonSet(ds *appsv1.DaemonSet, expectedGeneration int64, forceRollout bool) (*appsv1.DaemonSet, bool, error) {
-	s.daemonSets = append(s.daemonSets, ds)
+	dsRollout := daemonSetRollout{forceRollout: forceRollout, daemonSet: ds}
```
jsafrane: I can see `forceRollout` only written, but it does not seem to be read anywhere in the test

gnufied: fixed

PR #85, opened 2020-01-31
Bug 1788916: Blocking owner deletion should prevent LocalObject from disappearing

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1788916

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -14,7 +14,6 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/k8sclient"
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -824,11 +824,12 @@ func addOwner(meta *metav1.ObjectMeta, cr *localv1.LocalVolume) {
...
-			Kind:       localv1.LocalVolumeKind,
-			Name:       cr.Name,
-			UID:        cr.UID,
-			Controller: &trueVal,
+			APIVersion:         localv1.SchemeGroupVersion.String(),
+			Kind:               localv1.LocalVolumeKind,
+			Name:               cr.Name,
+			UID:                cr.UID,
+			Controller:         &trueVal,
+			BlockOwnerDeletion: &trueVal,
```
jsafrane: Kubernetes docs says

> 	// If true, AND if the owner has the "foregroundDeletion" finalizer, then
	// the owner cannot be deleted from the key-value store until this
	// reference is removed.
	// Defaults to false.

I don't see `foregroundDeletion` set anywhere in the code.

And what is the expected outcome? The documentation above says the user cannot delete the owner (LocalVolume) without deleting owned objects (DaemonSets, ConfigMaps and whatnot), which does not make much sense.

Or is the documentation wrong?

Issue #87, opened 2020-02-04
no matches for kind "LocalVolume" in version "local.storage.openshift.io/v1"


achuzhoy:
Tried to follow the instructions here:
https://github.com/openshift/local-storage-operator/blob/master/docs/deploy-with-olm.md
oc create -f create-cr.yaml fails with
error: unable to recognize "create-cr.yaml": no matches for kind "LocalVolume" in version "local.storage.openshift.io/v1"

achuzhoy:
This was attempted on a cluster version: 4.3.0-0.nightly-2020-01-29-050742

cdvultur:
What is the procedure to install local-storage-operator on old versions ?

jnpacker:
Let me take a look at the Mongodb chart and see if I can track this down.

gnufied:
Instructions for OCP installation is here - https://docs.openshift.com/container-platform/4.2/storage/persistent-storage/persistent-storage-local.html

The docs you have linked is for internal development and testing.

ohochman:
this issue is preventing us from automating the ACM installation, as it required to add the storage from the UI..   according to https://docs.openshift.com/container-platform/4.2/storage/persistent-storage/persistent-storage-local.html 

Questions: 
(1) Is the link above it the right way to deploy MCM on OCP BM ? 
(2) Is there a way to add the storage in another way not from UI? 
(3) does the local storage is a mandatory requirement for the installation to start?

gnufied:
You can also install operator from official redhat channels by using `oc create -f https://github.com/openshift/local-storage-operator/blob/master/examples/olm/redhat-src-catalog.yaml`. Just make sure that channel in https://github.com/openshift/local-storage-operator/blob/master/examples/olm/redhat-src-catalog.yaml#L21 matches version of OCP installing the operator on.

There is a docs bug filed to document installation process from CLI - https://bugzilla.redhat.com/show_bug.cgi?id=1800649

I dunno what you mean by ACM installation but please open BZ against Redhat if you need help.

ElCoyote27:
I'm getting the same error on 4.5.8 specifically (following https://github.com/red-hat-storage/ocs-training/blob/master/ocp4ocs4/ocs4-install-no-ui.md)

This is a fresh 4.5.8 cluster:
```
$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.5.8     True        False         3h51m   Cluster version is 4.5.8
[...]

PR #88, opened 2020-02-05
Warn about installation on OCP from in-tree docs

gnufied:
We should warn people from using docs stored in github repo, so as we do not have issues like https://github.com/openshift/local-storage-operator/issues/87

### DIFF
--- a/docs/deploy-with-olm.md
+++ b/docs/deploy-with-olm.md
@@ -2,7 +2,7 @@
 
 The local-storage Operator is avaialble in OLM, this makes it easy to consume local storage on your OCP cluster.
 
-**Warning: These docs are for internal development and testing. Use https://docs.openshift.com/container-platform/4.2/storage/persistent-storage/persistent-storage-local.html docs for installation on OCP**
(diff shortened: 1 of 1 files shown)


docs/deploy-with-olm.md
```diff
@@ -2,6 +2,8 @@
 
 The local-storage Operator is avaialble in OLM, this makes it easy to consume local storage on your OCP cluster.
 
+**Warning: These docs are for internal development and testing. Use https://docs.openshift.com/container-platform/4.2/storage/persistent-storage/persistent-storage-local.html docs for installation on OCP**
```
jsafrane: I wonder how stable is `latest` in this: https://docs.openshift.com/container-platform/latest/storage/persistent-storage/persistent-storage-local.html
@huffmanca, are there any guarantees about link names?

huffmanca: `latest` should be acceptable here - it will redirect to the latest published version. If it directs to a password-protected unpublished version, then that's a bug.

gnufied: fixed

PR #91, opened 2020-02-14
Bug 1803215:Prevent deletion of children objects owned by LocalVolume object

gnufied:
Fix removal of children objects.


bparees:
skipping this one for approval this week due to bug severity and the number of other changes already accepted.

PR #100, opened 2020-03-27
Bug 1781502: Updating LSO to use digest for 4.3

huffmanca:
We currently use digests for 4.4 and 4.5; however, because we specify the versions instead of relying on the `olm.skipRange` to subscribe to the correct channel. Changing this to be `latest` should allow ART to correctly use the digest instead of the tag.

### DIFF
--- a/manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
+++ b/manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
@@ -29,7 +29,7 @@ metadata:
       ]
     categories: Storage
     capabilities: Full Lifecycle
-    containerImage: quay.io/openshift/origin-local-storage-operator:4.3.0
+    containerImage: quay.io/openshift/origin-local-storage-operator:latest
     support: Red Hat
     repository: https://github.com/openshift/local-storage-operator
     createdAt: "2019-08-14T00:00:00Z"
--- a/manifests/4.4/local-storage-operator.v4.4.0.clusterserviceversion.yaml
+++ b/manifests/4.4/local-storage-operator.v4.4.0.clusterserviceversion.yaml
@@ -29,7 +29,7 @@ metadata:
       ]
     categories: Storage
     capabilities: Full Lifecycle
-    containerImage: quay.io/openshift/origin-local-storage-operator:4.4.0
+    containerImage: quay.io/openshift/origin-local-storage-operator:latest
     support: Red Hat
     repository: https://github.com/openshift/local-storage-operator
     createdAt: "2019-08-14T00:00:00Z"


manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml
```diff
@@ -29,11 +29,12 @@ metadata:
       ]
     categories: Storage
     capabilities: Full Lifecycle
-    containerImage: quay.io/openshift/origin-local-storage-operator:4.2.0
+    containerImage: quay.io/openshift/origin-local-storage-operator:latest
     support: Red Hat
     repository: https://github.com/openshift/local-storage-operator
     createdAt: "2019-08-14T00:00:00Z"
     description: Configure and use local storage volumes in kubernetes and Openshift
+    olm.skipRange: '>=4.2.0 <4.3.0'
```
gnufied: A 4.2 manifest in 4.3 branch should have no effect IMO on build process. It is only kept for keeping build scripts happy when transition between branches happen.

manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
```diff
@@ -29,7 +29,7 @@ metadata:
       ]
     categories: Storage
     capabilities: Full Lifecycle
-    containerImage: quay.io/openshift/origin-local-storage-operator:4.3.0
+    containerImage: quay.io/openshift/origin-local-storage-operator:latest
```
gnufied: I thought this is only for display purpose in the UI. The actual image version is mentioned below which does get replaced by the ART pipeline.

manifests/4.2/local-storage-operator.v4.2.0.clusterserviceversion.yaml
huffmanca: Understood. I can remove this reference here.

manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml
gnufied: This is still a good idea to do but this may not fix the bug that was reported.

huffmanca: We discussed this on slack. I'm reaching out to the reporter of the issue to confirm that the images are still correct, and that it's simply the `containerImage` reporting the tag instead of the digest.

I'm leaving this as is for now, as the reported issue involves fetching the `containerImage` and it reporting an incorrect tag.

gnufied:
make it depend on https://bugzilla.redhat.com/show_bug.cgi?id=1770183 BZ? @huffmanca

PR #105, opened 2020-04-13
Bug 1822213: Include version in diskmaker and provisioner configmap

huffmanca:
It was noticed in [BZ1822213](https://bugzilla.redhat.com/show_bug.cgi?id=1822213) that the diskmaker and provisioner pods are not updated when the local storage operator is.

This PR adds the controller version into the configmaps. When the version changes the diskmaker and provisioner pods will be redeployed.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"os"
 	"sync"
@@ -36,6 +38,7 @@ type Handler struct {
 	diskMakerConfigName   string
 	lock                  sync.Mutex
 	apiClient             apiUpdater
+	controllerVersion     string
 }
 
 type localDiskData map[string]map[string]string
@@ -64,6 +67,8 @@ const (
 	provisionerImageEnv = "PROVISIONER_IMAGE"
 
 	localVolumeFinalizer = "storage.openshift.com/local-volume-protection"
+
+	specHashAnnotation = "operator.openshift.io/spec-hash"
 )
 
 // NewHandler returns a controller handler
@@ -583,6 +588,10 @@ func (h *Handler) syncDiskMakerDaemonset(cr *localv1.LocalVolume, forceRollout b
 	ds := h.generateDiskMakerDaemonSet(cr)
 	dsName := ds.Name
 	generation := getExpectedGeneration(cr, ds)
+
+	// See if we need to roll out the pods based on the DaemonSet hash
+	forceRollout = h.checkDaemonSetHash(ds, forceRollout)
+
 	ds, _, err := h.apiClient.applyDaemonSet(ds, generation, forceRollout)
 	if err != nil {
 		return nil, fmt.Errorf("error applying diskmaker daemonset %s: %v", dsName, err)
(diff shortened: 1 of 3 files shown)


gnufied:
@huffmanca we should do same thing that we decide in https://github.com/openshift/aws-ebs-csi-driver-operator/pull/32 . So currently the problem is when LSO is restarted after an upgrade, it *does* receive new version of diskmaker and static-provisioner images but `Apply*` functions don't work based on that and hence they are not rolled out.

huffmanca:
This should now be updated to include a hash inside the DaemonSet annotations. This hash is compared against any existing values to determine if we should force a rollout or not.

pkg/controller/controller.go
```diff
@@ -594,7 +613,21 @@ func (h *Handler) syncProvisionerDaemonset(cr *localv1.LocalVolume, forceRollout
...
+	existingDS, err := h.apiClient.getDaemonSet(ds.Namespace, dsName)
+	if err != nil {
+		klog.Info("Error getting existing provisioner DaemonSet: %v", err)
+	}
+	existingAnnotations := existingDS.ObjectMeta.Annotations
+	if existingAnnotations == nil ||
+		existingAnnotations[specHashAnnotation] != ds.ObjectMeta.Annotations[specHashAnnotation] {
+		forceRollout = true
+	}
+
```
gnufied, line 629: Have we tried pulling latest version of library-go which automatically does this - https://github.com/openshift/aws-ebs-csi-driver-operator/pull/49/files ?

gnufied: Also if we can't pull library-go then we should at least move this code that duplicates bunch of code in a function of its own and then write some tests.

huffmanca: As discussed offline - there are build issues after updating to the latest library-go. I've created a method that does this check.

pkg/controller/controller.go
```diff
@@ -811,12 +830,35 @@ func (h *Handler) generateDiskMakerDaemonSet(cr *localv1.LocalVolume) *appsv1.Da
...
 	addOwner(&ds.ObjectMeta, cr)
 	addOwnerLabels(&ds.ObjectMeta, cr)
 	return ds
 }
 
+func (h *Handler) checkDaemonSetHash(ds *appsv1.DaemonSet) bool {
+	forceRollout := false
+
+	existingDS, err := h.apiClient.getDaemonSet(ds.Namespace, ds.Name)
+	if err != nil {
```
gnufied, line 849: If there was an error fetching daemonset could it not be nil and taking annotations from a nil object will cause nil pointer dereference?

pkg/controller/controller.go
```diff
@@ -583,6 +588,10 @@ func (h *Handler) syncDiskMakerDaemonset(cr *localv1.LocalVolume, forceRollout b
 	ds := h.generateDiskMakerDaemonSet(cr)
 	dsName := ds.Name
 	generation := getExpectedGeneration(cr, ds)
+
+	// See if we need to roll out the pods based on the DaemonSet hash
+	forceRollout = h.checkDaemonSetHash(ds)
```
gnufied, line 593: This variable here overrides `forceRollout` parameter supplied to the function. This will create problems in cases where daemonset spec does not have any change but `LocalVolume` object has changed.

pkg/controller/controller.go
huffmanca: I updated this to return if an error is detected. That should prevent any nil pointer dereferences.

If no existing DaemonSet can be found, then the sync* methods should continue with their previous behavior, as `daemonSetUpdated` will be false.

pkg/controller/controller.go
huffmanca: This has been updated to be its own `bool`. If the new `daemonSetUpdated` is true, then `forceRollout` will be updated as well; otherwise, `forceRollout` keeps its original value.

gnufied: This does mean that we did not had had adequate tests to cover these scenarios. Can we make sure that this is covered?

huffmanca: I moved this logic into `checkDaemonSetHash()`, and updated `TestCheckDaemonSetHash` to examine a combination of `forceRollout` and updated daemon sets. Let me know your thoughts on the current approach.

gnufied:
@huffmanca can we cherry pick this to 4.3 and 4.4 plz?

PR #110, opened 2020-04-21
 Auto provision

rohantmp:
Implementation of `LocalVolumeSet`

Components:
- In the operator's manager:
    - nodedaemon controller: reconciles daemonsets, provisionerconfigmap
    - localvolumeset controller: reconciles storageclass, updates status
- In the diskmaker's manager:
    - localvolumeset controller: creates symlinks

TODO:
- [x] fix non-csv yamls
- [x] unit tests for utility methods
- [x] mocked unit tests for localvolumeset controller
- [x] e2e  functional tests for localvolumeset


pkg/internal/diskutil.go
```diff
@@ -0,0 +1,224 @@
...
+	return false, nil
+}
+
+// ListBlockDevices using the lsblk command
+func ListBlockDevices() ([]BlockDevice, []string, error) {
+	var output bytes.Buffer
+	var blockDevices []BlockDevice
+
+	columns := "NAME,ROTA,TYPE,SIZE,MODEL,VENDOR,RO,RM,STATE,FSTYPE"
+	args := []string{"--pairs", "-b", "-o", columns}
```
sp98, line 160: There is an option for using `--json`.  Could that be more useful here ?
`lsblk --json --output  "SIZE,ROTA,RO,TYPE,PKNAME,NAME,VENDOR,ROTA,FSTYPE,MODEL"`

rohantmp: won't be available in lsblk until we start using rhel8 based images

rohantmp: I started with that, and the rewrote it :grimacing:

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,134 @@
...
+	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
+	// StorageClassName to use for set of matched devices
+	StorageClassName string `json:"storageClassName"`
+	// VolumeMode determines whether the PV created is Block or Filesystem. By default it will
+	// be block
+	// + optional
+	// MinDeviceCount is the minumum number of devices that needs to be detected per node.
+	// If the match devices are less than the minCount specified then no PVs will be created.
+	// +optional
+	MinDeviceCount *int32 `json:"minDeviceCount"`
```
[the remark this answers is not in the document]
rohantmp: I'll remove it and it should be easy enough to add back as needed.

deploy/crds/local.storage.openshift.io_v1alpha1_localvolumeset_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: local.storage.openshift.io/v1alpha1
+kind: LocalVolumeSet
+metadata:
+  name: example-localvolumeset
+spec:
+  # Add fields here
+  size: 3
```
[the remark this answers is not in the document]
rohantmp: generated file. removed

cmd/manager/main.go
```diff
@@ -96,6 +96,7 @@ func main() {
 	options := manager.Options{
 		Namespace:          namespace,
 		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		LeaderElection:     false,
 	}
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,134 @@
...
+	Vendors []string `json:"vendors"`
+}
+
+// LocalVolumeSetSpec defines the desired state of LocalVolumeSet
+type LocalVolumeSetSpec struct {
+	// Nodes on which the automatic detection policies must run.
+	// +optional
+	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
+	// StorageClassName to use for set of matched devices
+	StorageClassName string `json:"storageClassName"`
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/internal/diskutil.go
```diff
@@ -0,0 +1,223 @@
...
+func (e IDPathNotFoundError) Error() string {
+	return fmt.Sprintf("IDPathNotFoundError: a symlink to  %q was not found in %q", e.DeviceName, DiskByIDDir)
+}
+
+// BlockDevice is the a block device as output by lsblk.
+// All the fields are lsblk columns.
+type BlockDevice struct {
+	Name   string `json:"name"`
+	Type   string `json:"type"`
+	Model  string `json:"mode,omitempty"`
```
sp98, line 36:
```suggestion
	Model  string `json:"model,omitempty"`
```

pkg/internal/diskutil.go
```diff
@@ -0,0 +1,223 @@
...
+	// convert to json and then Marshal.
+	outputMapList := make([]map[string]interface{}, 0)
+	rowList := strings.Split(output.String(), "\n")
+RowLoop:
+	for _, row := range rowList {
+		if len(strings.Trim(row, " ")) == 0 {
+			break RowLoop
+		}
+		outputMap := make(map[string]interface{})
+		// split by `" ` to avoid splitting on spaces in MODEL,VENDOR
```
sp98, line 177: Can we also trim the white spaces in model and vendor names.

pkg/controller/localvolumeset/reconcile.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, storageClass, func() error {
+		if storageClass.CreationTimestamp.IsZero() {
+			storageClass.Provisioner = "kubernetes.io/no-provisioner"
+			storageClass.ReclaimPolicy = &deleteReclaimPolicy
+			storageClass.VolumeBindingMode = &firstConsumerBinding
+			storageClass.ObjectMeta.Labels = map[string]string{
+				util.OwnerNameLabel:      lvs.GetName(),
```
gnufied, line 120: non-namespaced objects can't hold referenced to namespaced objects - https://github.com/openshift/local-storage-operator/pull/114/files it results in hard to track bugs.

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -0,0 +1,76 @@
...
+	desiredDm := generateDiskMakerDaemonset(request, tolerations, ownerRefs, nodeSelector)
+	opResult, err = createOrUpdateDaemonset(r.client, desiredDm)
+	if err != nil {
+		return reconcile.Result{}, err
+	} else if opResult == controllerutil.OperationResultUpdated || opResult == controllerutil.OperationResultCreated {
+		r.reqLogger.Info("daemonset changed", "daemonset.Name", desiredDm.GetName(), "op.Result", opResult)
+	}
+
+	desiredDm = generateLocalProvisionerDaemonset(request, tolerations, ownerRefs, nodeSelector)
+	opResult, err = createOrUpdateDaemonset(r.client, desiredDm)
```
gnufied, line 68: We should try and use one set of functions behind a common interface from this operator.

pkg/controller/localvolumeset/reconcile.go
```diff
@@ -0,0 +1,133 @@
...
+	// this way, there can be one daemonset for all LocalVolumeSets
+
+	err = r.syncStorageClass(instance)
+	if err != nil {
+		r.reqLogger.Error(err, "failed to sync storageclass")
+		return reconcile.Result{}, err
+	}
+
+	r.reqLogger.Info("updating status")
+	err = r.updateStatus(instance)
```
gnufied, line 77: so if I understand this correctly, we aren't setting Conditions field in status anywhere?

pkg/diskmaker/controllers/lvset/add.go
```diff
@@ -0,0 +1,55 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_localvolumeset")
+
+// Add adds a new nodeside lvset controller to mgr
+func Add(mgr manager.Manager) error {
+	r := &ReconcileLocalVolumeSet{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	// Create a new controller
+	c, err := controller.New("localvolumeset-controller", mgr, controller.Options{Reconciler: r})
```
gnufied, line 22: same name for both controller on the node and controller on the control-plane? in fact, I would like us to move away from name "controller" for components that are running on the node

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -0,0 +1,265 @@
...
+	cm := &corev1.ConfigMap{}
+	// TOODO(rohan) export/import name
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: nodedaemon.ProvisionerConfigMapName, Namespace: request.Namespace}, cm)
+	if err != nil {
+		reqLogger.Error(err, "could not get provisioner configmap")
+		return reconcile.Result{}, err
+	}
+
+	provisionerConfig := staticProvisioner.ProvisionerConfiguration{}
+	staticProvisioner.ConfigMapDataToVolumeConfig(cm.Data, &provisionerConfig)
```
gnufied, line 82: lets move the device discovery and symlinking code into its own function rather than having everything here.

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -0,0 +1,265 @@
...
+		reqLogger.Error(err, "could not list block devices", "lsblk.BadRows", badRows)
+		return reconcile.Result{}, err
+	} else if len(badRows) > 0 {
+		reqLogger.Error(fmt.Errorf("bad rows"), "could not parse all the lsblk rows", "lsblk.BadRows", badRows)
+	}
+
+	validDevices := make([]internal.BlockDevice, 0)
+
+	// get valid devices
+DeviceLoop:
```
gnufied, line 102: can we please remove these goto style conditionals?

pkg/controller/nodedaemon/daemonsets.go
```diff
@@ -0,0 +1,254 @@
...
+	out.ObjectMeta.Annotations = in.ObjectMeta.Annotations
+	out.ObjectMeta.OwnerReferences = in.ObjectMeta.OwnerReferences
+
+	out.Spec.Template.ObjectMeta.Annotations = in.Spec.Template.ObjectMeta.Annotations
+	out.Spec.Template.Spec = in.Spec.Template.Spec
+
+	out.Spec.UpdateStrategy = in.Spec.UpdateStrategy
+	out.Spec.MinReadySeconds = in.Spec.MinReadySeconds
+	out.Spec.RevisionHistoryLimit = in.Spec.RevisionHistoryLimit
+	return
```
rohantmp, line 43: copying @jsafrane 's [comment](https://github.com/openshift/local-storage-operator/pull/123#discussion_r456509827) here as this is where the change is. #123 is based on this.



I checked the way you apply DaemonSet and I can immediately spot an issue: what if user (or another operator / controller / whatever) changes Spec.Selector? This code does not fix the selector back to the state the operator wants it. This may result in unexpected and very hard to debug behavior of the whole LSO.

Not mentioning that you're abusing the API server by calling Update even when there may be no actual update necessary.

Of course, you can add the selector to things the operator replaces. But explicitly listing fields you're going to fix is very brittle. In the future, Kubernetes may introduce a new DaemonSet.Spec field that the operator does not know about and it will not fix it. And it may break the operator again.

What is 100% safe and future proof is to remember the last Generation of the DaemonSet that LSO created / updated. When the operator gets an event that the DaemonSet has updated, it can compare the current Generation with the last correct Generation and immediately see that someone / something has changed DaemonSet.Spec and overwrite the whole Spec. And that's why we're tracking generations of children in LSO Status.

rohantmp: > I checked the way you apply DaemonSet and I can immediately spot an issue: what if user (or another operator / controller / whatever) changes Spec.Selector? This code does not fix the selector back to the state the operator wants it. This may result in unexpected and very hard to debug behavior of the whole LSO.

`DaemonSetSpec.Spec.Selector` is an immutable field.

> Not mentioning that you're abusing the API server by calling Update even when there may be no actual update necessary.

[Not true](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L216):
```
	if equality.Semantic.DeepEqual(existing, obj) {
		return OperationResultNone, nil
	}
```
> In the future, Kubernetes may introduce a new DaemonSet.Spec field that the operator does not know about and it will not fix it. And it may break the operator again.

we'd have to implement using that field in code too.

> What is 100% safe and future proof is to remember the last Generation of the DaemonSet that LSO created / updated. When the operator gets an event that the DaemonSet has updated, it can compare the current Generation with the last correct Generation and immediately see that someone / something has changed DaemonSet.Spec and overwrite the whole Spec. And that's why we're tracking generations of children in LSO Status.

Here,  that is taken care of by the watch mechanism. It ensures that the reconciler runs every time it starts or sees a cache update (by tracking the generation).

pkg/controller/nodedaemon/reconcile.go
rohantmp: createOrUpdateDaemonset is a utility method that  a common interface underneath: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/client#Client

the struct that implements that interface is a `DelegatingClient`
```
type DelegatingClient struct {
    Reader
    Writer
    StatusClient
}
```
This is beneficial because:

> DelegatingClient forms a Client by composing separate reader, writer and statusclient interfaces. This way, you can have an Client that reads from a cache and writes to the API server.

pkg/controller/util/common.go
```diff
@@ -0,0 +1,50 @@
+package util
+
+import "os"
+
+const (
+	defaultDiskMakerImageVersion = "quay.io/openshift/origin-local-storage-diskmaker"
+	defaultProvisionImage        = "quay.io/openshift/origin-local-storage-static-provisioner"
+	localDiskLocation            = "/mnt/local"
```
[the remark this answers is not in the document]
rohantmp: I'll try moving it to the same directory and see if there are any issues.

pkg/diskmaker/controllers/lvset/add.go
rohantmp: `controller` here is the name of the `controller-runtime` struct.

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: Though they are not goto's, jumping out of nested loops is actually one situation where goto is not considered harmful :)
These labeled `continue`s help write more readable code i.e. 'skip this device' instead of something like:
```
	for _, blockDevice := range blockDevices {
		shouldSkip := false
		for name, filter := range filterMap { // inner loop
			if !notMatch {
				shouldSkip = true
				break
			}
		}

		if shouldSkip {
			continue
		}
		for name, filter := range matcherMap { // inner loop
			if !notMatch {
				shouldSkip = true
				break
			}
		}
		if shouldSkip {
			continue
		}
		//something else
	}

```

pkg/internal/diskutil.go
rohantmp: I think we should include whatever is between the quoutes in the output. Either way, the matcher only depends on whether the included value is a substring of this value.

pkg/internal/diskutil.go
```diff
@@ -0,0 +1,227 @@
...
+
+// BlockDevice is the a block device as output by lsblk.
+// All the fields are lsblk columns.
+type BlockDevice struct {
+	Name   string `json:"name"`
+	Type   string `json:"type"`
+	Model  string `json:"mode,omitempty"`
+	Vendor string `json:"vendor,omitempty"`
+	State  string `json:"state,omitempty"`
+	FSType string `json:"fstype"`
```
sp98, line 40:
```suggestion
	FSType string `json:"fsType"`
```

pkg/diskmaker/controllers/lvset/add.go
gnufied: I meant `localvolumeset-controller` string

gnufied: Also, what I meant by moving away from controller for components running on node is - if you open `cmd/diskmaker` folder, you will notice two files - "diskmaker.go" and "controller.go". What is "controller.go" supposed to mean in that context? How is it different from "diskmaker.go"?

pkg/controller/nodedaemon/daemonsets.go
jsafrane: > if equality.Semantic.DeepEqual(existing, obj) {
>		return OperationResultNone, nil
>	}

If you replace `obj.Spec.Template`, then it will always return that the objects are different because of defaulting in API server.

> Here, that is taken care of by the watch mechanism. It ensures that the reconciler runs every time it starts or sees a cache update (by tracking the generation).

I am not sure I understand. Where LSO tracks generation of the newly introduced DaemonSet?

gnufied:
@rohantmp Couple of bugs:

1. Even though `nodeSelector` is marked as an optional field, not specifying it results in:

```
{"level":"info","ts":1595254918.7468674,"logger":"localvolumesetdaemon-controller","msg":"provisioner configmap changed","Request.Namespace":"default","Request.Name":""}                                            {"level":"error","ts":1595254918.7535434,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"localvolumesetdaemon-controller","request":"default/","error":"DaemonSet.apps \"diskmaker-manager\" is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms: Required value: must have at least one node selector term","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift/local-storage-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/src/github.com/openshift/local-storage-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/src/github.com/openshift/local-storage-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/go/src/github.com/openshift/local-storage-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/go/src/github.com/openshift/local-storage-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/src/github.com/openshift/local-storage-operator/vendor/k8s.io/a
```

2. Even if `deviceInclusionSpec` is marked as optional, not specifying it in CR cahses diskmaker pod to crash.
3. I don't recall if we agreed that all devices selected by LSO must have a symlink in `/dev/disk/by-id` because this would exclude any disks attached to a VM by libvirt tools.  
4. There is no error reporting of any kind. There is no events, no error conditions present on `LocalVolumeSet` object type.

cc @travisn

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -0,0 +1,76 @@
...
+
+	desiredDm := generateDiskMakerDaemonset(request, tolerations, ownerRefs, nodeSelector)
+	opResult, err = createOrUpdateDaemonset(r.client, desiredDm)
+	if err != nil {
+		return reconcile.Result{}, err
+	} else if opResult == controllerutil.OperationResultUpdated || opResult == controllerutil.OperationResultCreated {
+		r.reqLogger.Info("daemonset changed", "daemonset.Name", desiredDm.GetName(), "op.Result", opResult)
+	}
+
+	desiredDm = generateLocalProvisionerDaemonset(request, tolerations, ownerRefs, nodeSelector)
```
gnufied, line 67: If there is a change in configmap (for example - `fsType` changed) then shouldn't the daemonset be rolled out?

pkg/diskmaker/controllers/lvset/reconcile.go
gnufied: ack. I think it is fine to use labels where it makes sense. Marking as resolved.

pkg/internal/diskutil.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	badRows := make([]string, 0)
+	// convert to json and then Marshal.
+	outputMapList := make([]map[string]interface{}, 0)
+	rowList := strings.Split(output.String(), "\n")
+RowLoop:
+	for _, row := range rowList {
+		if len(strings.Trim(row, " ")) == 0 {
+			break RowLoop
```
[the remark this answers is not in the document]
rohantmp: removing, it's an artifact

pkg/diskmaker/controllers/lvset/matcher.go
```diff
@@ -0,0 +1,167 @@
...
+	inVendorList             = "inVendorList"
+	inModelList              = "inModelList"
+)
+
+// maps of function identifier (for logs) to filter function.
+// These are passed the localv1alpha1.DeviceInclusionSpec to make testing easier,
+// but they aren't expected to use it
+// they verify that the device itself is good to use
+var filterMap = map[string]func(internal.BlockDevice, localv1alpha1.DeviceInclusionSpec) (bool, error){
+	notReadOnly: func(dev internal.BlockDevice, spec localv1alpha1.DeviceInclusionSpec) (bool, error) {
```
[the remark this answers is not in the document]
rohantmp: I didn't have it initially, but keeping it generic makes it much easier to unit test.

pkg/controller/nodedaemon/reconcile.go
rohantmp: DaemonSet has the `RollingUpdate` strategy set.

pkg/controller/localvolumeset/reconcile.go
rohantmp: Had already removed in unpushed version :grimacing:

pkg/controller/add_nodedaemon.go
```diff
@@ -0,0 +1,10 @@
+package controller
+
+import (
+	"github.com/openshift/local-storage-operator/pkg/controller/nodedaemon"
+)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	AddToManagerFuncs = append(AddToManagerFuncs, nodedaemon.AddDaemonReconciler)
```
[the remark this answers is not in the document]
rohantmp: It's an artifact of operator-sdk generation. I think the idea is that different controllers have full file separation this way.

It definitely makes it easier to separate commits by component, but I don't feel to strongly about this. I'll leave it up to you.

pkg/controller/nodedaemon/daemonsets.go
rohantmp: > If you replace obj.Spec.Template, then it will always return that the objects are different because of defaulting in API server.

Ah, I see, I'll investigate. Does apply not also hit the API server?

 > I am not sure I understand. Where LSO tracks generation of the newly introduced DaemonSet?

It doesn't but the reconciler runs for every generation change, but I see what you mean. I will look into implementing generation tracking for the purpose of bubbling up errors.

pkg/controller/nodedaemon/reconcile.go
gnufied: I am not sure I understand. if configmap changed but there was no change at al in daemonset then it will cause daemonset to be not updated. How does `RollingUpdate` help with that?

.gitignore
```diff
@@ -4,3 +4,4 @@ diskmaker
 !diskmaker/
 cover.out
 _output
+.vscode/
```
gnufied, line 7: nit: strictly speaking this change can be made to your global `$HOME/.gitignore` file rather than in this file.

pkg/controller/add_nodedaemon.go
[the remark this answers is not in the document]
gnufied: I would prefer these statements in  a single file please. code generation by a tool should not affect our own code organization.

pkg/diskmaker/controllers/lvset/matcher.go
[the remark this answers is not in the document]
gnufied: I do not understand why will a function receive an argument that it never uses. How does it make unit testing easier? You can just choose to not supply that argument..

pkg/controller/nodedaemon/daemonsets.go
jsafrane: Yes, `ApplyDaemonSet` in library-go hits API server, and it mostly does Get, not Update, which is not that invasive.

rohantmp:
pushed e2e fixes and PV creation locking

will push clean up and comment on each addressed review comment as fixes are pushed

gnufied:
@rohantmp can you push a fix for `/dev/disk/by-id` thing, so as it can work in my local environment.

rohantmp:
> @rohantmp can you push a fix for /dev/disk/by-id thing, so as it can work in my local environment.

It should be working now

https://github.com/rohantmp/local-storage-operator/blob/autoProvision/pkg/diskmaker/controllers/lvset/reconcile.go#L223

pkg/diskmaker/controllers/lvset/matcher.go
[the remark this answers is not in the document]
rohantmp: In this case: https://github.com/rohantmp/local-storage-operator/blob/8b2aa777ac8416bdd25ba103e50fcb8947430737/pkg/diskmaker/controllers/lvset/matcher_test.go#L540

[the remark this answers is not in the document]
rohantmp: But yeah, it was mostly a refactoring concern and I can remove it.

pkg/controller/util/common.go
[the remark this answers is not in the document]
rohantmp: done

rohantmp:
Unit test failed because fake client wasn't vendored. Only thing I changed was `go mod tidy && go mod vendor`

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// VolumeMode determines whether the PV created is Block or Filesystem. By default it will
+	// be block
+	// + optional
+	// Maximum number of Devices that needs to be detected per node.
+	// +optional
+	MaxDeviceCount *int32 `json:"maxDeviceCount,omitempty"`
+	// +optional
+	VolumeMode localv1.PersistentVolumeMode `json:"volumeMode,omitempty"`
```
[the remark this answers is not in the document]
gnufied: This is contrary to how we default elsewhere. The default IMO should be filesystem.

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,122 @@
...
+	StorageClassName string `json:"storageClassName"`
+	// Maximum number of Devices that needs to be detected per node.
+	// +optional
+	MaxDeviceCount *int32 `json:"maxDeviceCount,omitempty"`
+	// VolumeMode determines whether the PV created is Block or Filesystem. By default it will
+	// be block
+	// +optional
+	VolumeMode localv1.PersistentVolumeMode `json:"volumeMode,omitempty"`
+	// FSType type to create when volumeMode is Filesystem
+	// Mandatory if volumeMode is filesystem.
```
gnufied, line 76: I think default is ext4 if unspecified. I do not think there is a way to validate this field one way or the other if `volumeModeis set"

pkg/diskmaker/diskmaker.go
```diff
@@ -180,8 +192,27 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
 
 	for storageClass, deviceArray := range deviceMap {
 		for _, deviceNameLocation := range deviceArray {
+			// get PV creation lock which checks for existing symlinks to this device
+			pvLock, pvLocked, existingSymlinks, err := internal.GetPVCreationLock(
+				deviceNameLocation.diskNamePath,
+				d.symlinkLocation,
```
gnufied, line 198: So this will cause a lock to be acquired on all of the specified  devices and they will only be released when the whole function exits. This is usually a bad idea and can cause deadlocks, we should acquire lock on one device at a time and release the lock immediately after processing of that device is done.

pkg/diskmaker/diskmaker.go
```diff
@@ -180,8 +192,27 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
...
+				if err != nil {
+					klog.Errorf("failed to unlock device: %+v", err)
+				}
+			}
+			defer unlockFunc()
+			if len(existingSymlinks) > 0 { // already claimed, fail silently
+				continue
+			} else if err != nil || !pvLocked { // locking failed for some other reasion
+				klog.Errorf("not symlinking, could not get lock: %v", err)
+				continue
```
gnufied, line 211: We should report errors as events in case we can't claim a device because it was previously symlinked.

pkg/diskmaker/diskmaker.go
```diff
@@ -203,14 +235,14 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
...
-
 			if symLinkErr != nil {
 				msg := fmt.Sprintf("error creating symlink %s: %v", symLinkPath, symLinkErr)
 				e := newEvent(ErrorFindingMatchingDisk, msg, deviceNameLocation.diskNamePath)
 				d.eventSync.report(e, d.localVolume)
 				klog.Errorf(msg)
 			}
-
+			// unlock before proceeding
+			unlockFunc()
```
gnufied, line 245: if any of those `continue` statements are encountered that that will cause the lock to be released at the end of the function.

pkg/diskmaker/diskmaker.go
```diff
@@ -180,8 +192,27 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
...
+			if len(existingSymlinks) > 0 { // already claimed, fail silently
+				continue
+			} else if err != nil || !pvLocked { // locking failed for some other reasion
+				klog.Errorf("not symlinking, could not get lock: %v", err)
+				continue
+			}
+
 			symLinkDirPath := path.Join(d.symlinkLocation, storageClass)
-			err := os.MkdirAll(symLinkDirPath, 0755)
+			err = os.MkdirAll(symLinkDirPath, 0755)
```
gnufied, line 215: Should we create this directory regardless of error above? Because without that, the code throws:

> E0727 19:27:58.085659       1 discovery.go:201] Error reading directory: open /mnt/local-storage/two: no such file or directory

pkg/internal/diskutil.go
```diff
@@ -0,0 +1,338 @@
...
+	StateSuspended = "suspended"
+	// DiskByIDDir is the path for symlinks to the device by id.
+	DiskByIDDir = "/dev/disk/by-id/"
+)
+
+var (
+	// DefaultLockTimeout to use for WaitForPVCreationLock
+	DefaultLockTimeout = time.Second * 30
+	// DefaultLockInterval to use for
+	DefaultLockInterval = time.Second
```
gnufied, line 29: is anything using these values?

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -0,0 +1,297 @@
...
+	validDevices := r.getValidDevices(reqLogger, lvset, blockDevices)
+
+	// process valid devices
+	var noMatch []string
+	for _, blockDevice := range validDevices {
+		devLogger := reqLogger.WithValues("Device.Name", blockDevice.Name)
+
+		// validate MaxDeviceCount
+		var alreadyProvisionedCount int
+		alreadyProvisionedCount, noMatch, err = getAlreadyProvisioned(symLinkDir, validDevices)
```
gnufied, line 111: Can this use `GetMatchingSymlinksInDirs` function?

pkg/controller/nodedaemon/reconcile.go
gnufied: does the new approach work? I tried updating `LocalVolumSet` object and nothing happens.

gnufied:
@rohantmp To avoid racing with kubelet and ensuring that we don't pick devices that could be used by regular pods, I thought there was a plan to only pick a device after certain timeout, so as we don't inadvertently pick such devices.

pkg/controller/localvolumeset/map.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// store a one to many association from storageClass to LocalVolumeSet,
+// so that one PV event can fan out requests to all LocalVolumeSets.
+type lvSetMapStore struct {
+	storageClassMap map[string]map[types.NamespacedName]struct{}
```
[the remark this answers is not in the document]
jsafrane: `sets.Strings` (or a new flavour for `NamespacedName` would be even better.

pkg/controller/nodedaemon/daemonset_test.go
```diff
@@ -0,0 +1,44 @@
...
+	mutateAggregatedSpec(
+		ds,
+		reconcile.Request{},
+		[]corev1.Toleration{},
+		[]metav1.OwnerReference{},
+		nodeSelector,
+		"",
+		"",
+	)
+	assert.NotNilf(t, ds.Spec.Template.Spec.Affinity, "DaemonSet affinity shold not be nil if nodeSelector is not nil")
```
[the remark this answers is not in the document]
rohantmp: +1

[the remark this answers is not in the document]
rohantmp: commited fix, will be in next push

pkg/controller/nodedaemon/daemonset_test.go
```diff
@@ -0,0 +1,44 @@
...
+	mutateAggregatedSpec(
+		ds,
+		reconcile.Request{},
+		[]corev1.Toleration{},
+		[]metav1.OwnerReference{},
+		nil,
+		"",
+		"",
+	)
+	assert.Nilf(t, ds.Spec.Template.Spec.Affinity, "DaemonSet affinity shold be nil if nodeSelector is nil")
```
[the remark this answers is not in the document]
rohantmp: commited fix, will be in next push

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: It also reports nonMatching symlinks

pkg/internal/diskutil.go
rohantmp: removed, commited fix, will be in next push

pkg/diskmaker/diskmaker.go
rohantmp: Added change to include unlockfunc before each of the exit points of the loop. Will be in next push
Will open bug to refactor based on entry in our triaging google sheet (as discussed)

test/e2e/localvolumeset_test.go
```diff
@@ -0,0 +1,334 @@
...
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	labelNodeRoleWorker = "node-role.kubernetes.io/worker"
+)
+
+func LocalVolumeSetTest(ctx *framework.Context, cleanupFuncs *[]cleanupFn) func(*testing.T) {
```
gnufied, line 27: Why do we need gomega? Most of the e2e worked fine without it. Coming from k8s - i would rather avoid it if I can.

rohantmp: I used it because gomega.Eventually is very useful. 
Not using the whole ginkgo+gomega, this is just an assertion utility on a `*testing.T`.

pkg/diskmaker/diskmaker.go
rohantmp: not related to my PR. This is existing functionality. PR scope is already huge.

pkg/controller/nodedaemon/add.go
```diff
@@ -0,0 +1,97 @@
...
+func AddDaemonReconciler(mgr manager.Manager) error {
+	r := &DaemonReconciler{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	// Create a new controller
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// The controller will ignore the name part of the enqueued request as
+	// every reconcile gathers multiple resources an acts on a few one-per-namespace obects.
```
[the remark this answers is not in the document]
rohantmp: fix commited, will be in next push

cmd/diskmaker/controller.go
```diff
@@ -0,0 +1,93 @@
...
+
+	diskmakerController "github.com/openshift/local-storage-operator/pkg/diskmaker/controllers"
+)
+
+func startController(cmd *cobra.Command, args []string) error {
+	klogFlags := flag.NewFlagSet("local-storage-diskmaker", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
+	flag.Set("alsologtostderr", "true")
+	flag.Parse()
+	// Use a zap logr.Logger implementation. If none of the zap
```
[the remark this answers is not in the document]
rohantmp: I think so, we needed another entry-point

cmd/diskmaker/main.go
```diff
@@ -0,0 +1,33 @@
...
+var managerCmd = &cobra.Command{
+	Use:   "lv-manager",
+	Short: "Used to start the controller-runtime manager that owns the LocalVolumeSet controller",
+	RunE:  startController,
+}
+var lvDaemonCmd = &cobra.Command{
+	Use:   "lv-controller",
+	Short: "Used to start an instance of the controller for a specific LocalVolume CR",
+	RunE:  startDiskMaker,
+}
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+	localv1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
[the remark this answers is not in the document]
rohantmp: fixed

test/e2e/localvolumeset_test.go
gnufied: hmm.. kinda a nit but nonetheless, I do not see value in bringing yet another dependency. we already had - https://github.com/stretchr/testify which provides a decent collection of assertions that we can use without having to import yet another matcher/assertion library.

In terms of style and coherency of code - it makes maintenance easier if one style is followed throughout the code. I would probably remove gomega completely if I were you, but that just may be me.

pkg/apis/local/v1alpha1/localvolumeset_types.go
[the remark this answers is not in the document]
rohantmp: commited fix, will be in next push

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// LocalVolumeSetSpec defines the desired state of LocalVolumeSet
+type LocalVolumeSetSpec struct {
+	// Nodes on which the automatic detection policies must run.
+	// +optional
+	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
+	// StorageClassName to use for set of matched devices
+	StorageClassName string `json:"storageClassName"`
+	// VolumeMode determines whether the PV created is Block or Filesystem. By default it will
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// be block
+	// + optional
+	// Maximum number of Devices that needs to be detected per node.
+	// +optional
+	MaxDeviceCount *int32 `json:"maxDeviceCount,omitempty"`
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/controller/add_nodedaemon.go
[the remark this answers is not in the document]
rohantmp: commited fix, will be in next push

manifests/4.5/local-storage-operator.v4.5.0.clusterserviceversion.yaml
```diff
@@ -215,6 +227,36 @@ spec:
...
+            - get
+            - watch
+          - apiGroups:
+            - ""
+            - storage.k8s.io
+            resources:
+            - configmaps
+            - storageclasses
+            - persistentvolumeclaims
+            - persistentvolumes
```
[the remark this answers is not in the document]
rohantmp: yes

pkg/controller/util/common.go
```diff
@@ -0,0 +1,50 @@
...
+
+	ProvisionerServiceAccount    = "local-storage-admin"
+	ProvisionerPVRoleBindingName = "local-storage-provisioner-pv-binding"
+	ProvisionerNodeRoleName      = "local-storage-provisioner-node-clusterrole"
+
+	LocalVolumeSetRoleName        = "local-storage-provisioner-cr-role"
+	LocalVolumeSetRoleBindingName = "local-storage-provisioner-cr-rolebinding"
+
+	DefaultPVClusterRole           = "system:persistent-volume-provisioner"
+	ProvisionerNodeRoleBindingName = "local-storage-provisioner-node-binding"
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/controller/localvolumeset/map.go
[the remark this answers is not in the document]
rohantmp: Per discussion, have  captured in a google sheet as an issue to fix during feature freeze.

FWIW, sets.String does it the same way in the backend and wraps it with methods

pkg/controller/localvolumeset/map.go
```diff
@@ -0,0 +1,68 @@
...
+	mux             sync.Mutex
+}
+
+func (l *lvSetMapStore) getLocalVolumeSets(storageClass string) []types.NamespacedName {
+	l.mux.Lock()
+	defer l.mux.Unlock()
+	if len(l.storageClassMap) < 1 {
```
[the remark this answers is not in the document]
rohantmp: this function is never actually used in more than one thread at a time, leaving as-is

pkg/controller/localvolumeset/reconcile.go
```diff
@@ -0,0 +1,133 @@
...
+	deleteReclaimPolicy := corev1.PersistentVolumeReclaimDelete
+	firstConsumerBinding := storagev1.VolumeBindingWaitForFirstConsumer
+	storageClass := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      lvs.Spec.StorageClassName,
+			Namespace: lvs.GetNamespace(),
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, storageClass, func() error {
```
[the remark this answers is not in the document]
rohantmp: addressed in comment [below](https://github.com/openshift/local-storage-operator/pull/110#discussion_r457167827) and per discussions, we are leaving this in.

pkg/controller/localvolumeset/reconcile.go
rohantmp: fixed

pkg/controller/nodedaemon/add.go
```diff
@@ -0,0 +1,97 @@
...
+
```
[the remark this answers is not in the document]
rohantmp: fixed

pkg/controller/nodedaemon/provisioner_config.go
```diff
@@ -0,0 +1,58 @@
...
+			FsType:     lvSet.Spec.FSType,
+			HostDir:    symlinkDir,
+			MountDir:   symlinkDir,
+			VolumeMode: string(lvSet.Spec.VolumeMode),
+		}
+		storageClassConfig[storageClassName] = mountConfig
+	}
+	// create or update
+	return controllerutil.CreateOrUpdate(context.TODO(), r.client, configMap, func() error {
+		if configMap.CreationTimestamp.IsZero() {
```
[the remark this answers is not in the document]
rohantmp: addressed above

pkg/controller/localvolumeset/reconcile.go
rohantmp: fixed

pkg/diskmaker/controllers/lvset/add.go
rohantmp: fixed

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: fixed

pkg/controller/nodedaemon/daemonsets.go
rohantmp: Solved by defining the daemonset mutateFuncs to not override defaults.

pkg/controller/localvolumeset/map.go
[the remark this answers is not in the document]
gnufied: sure - it looks very ugly to read code like - `map[types.NamespacedName]struct{}{name: struct{}{}}` , personally for me - it takes a minute to process what is going on here. :-)

pkg/internal/diskutil.go
rohantmp: fixed

pkg/diskmaker/controllers/lvset/matcher.go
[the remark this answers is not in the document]
rohantmp: Fixed. It now accepts a pointer to the spec, but I kept it with the same signature as matcherFunc so that they could be unit tested with the same funcs

pkg/controller/nodedaemon/reconcile.go
rohantmp: fixed: was setting daemonset.metadata.annotations instead of daemonset.spec.template.metadata.annotations

.gitignore
rohantmp: removed

pkg/apis/local/v1alpha1/localvolumeset_types.go
rohantmp: fixed

pkg/diskmaker/diskmaker.go
rohantmp: fixed

pkg/diskmaker/diskmaker.go
rohantmp: added

rohantmp:
Addressed all comments :partying_face:

gnufied:
lets refactor and fix things in a follow up.

/lgtm

PR #116, opened 2020-06-01
Bug 1779545: Use double quote for matching skiprange

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1779545


gnufied:
cc @openshift/sig-storage

PR #117, opened 2020-06-01
Fix quote around version

gnufied:
This fixes version mismatch when comparing skipRange string.


jsafrane:
/lgtm
however, is master 4.6 now?

gnufied:
@jsafrane yeah we will have to update the manifest to follow 4.6 channel.

PR #118, opened 2020-06-01
Bug 1850090: [4.4] Use double quotes for skipRange

gnufied:
so as it can match same string in art.yaml


jsafrane:
But it needs a valid BZ to get merged

PR #123, opened 2020-06-04
Auto discovery of devices

sp98:
Description:
Allows discovery of potentially usable local disks on chosen nodes.

Design PR: https://github.com/openshift/enhancements/pull/237

Pending Items:

- [x] merging lvset and discovery e2e set


pkg/controller/nodedaemon/daemonsets.go
```diff
@@ -0,0 +1,254 @@
...
+	out.ObjectMeta.Annotations = in.ObjectMeta.Annotations
+	out.ObjectMeta.OwnerReferences = in.ObjectMeta.OwnerReferences
+
+	out.Spec.Template.ObjectMeta.Annotations = in.Spec.Template.ObjectMeta.Annotations
+	out.Spec.Template.Spec = in.Spec.Template.Spec
+
+	out.Spec.UpdateStrategy = in.Spec.UpdateStrategy
+	out.Spec.MinReadySeconds = in.Spec.MinReadySeconds
+	out.Spec.RevisionHistoryLimit = in.Spec.RevisionHistoryLimit
+	return
```
jsafrane, line 43: I checked the way you apply DaemonSet and I can immediately spot an issue: what if user (or another operator / controller / whatever) changes Spec.Selector? This code does not fix the selector back to the state the operator wants it. This may result in unexpected and **very** hard to debug behavior of the whole LSO.

Not mentioning that you're abusing the API server by calling Update even when there may be no actual update necessary.

Of course, you can add the selector to things the operator replaces. But explicitly listing fields you're going to fix is very brittle. In the future, Kubernetes may introduce a new DaemonSet.Spec field that the operator does not know about and it will not fix it. And it may break the operator again.

What is 100% safe and future proof is to remember the last Generation of the DaemonSet that LSO created / updated. When the generation of the object changes, the operator can compare the current Generation with the last correct Generation and immediately see that someone / something has changes DaemonSet.Spec and overwrite the whole spec. And that's why we're tracking generations of children in LSO Status.

rohantmp: Copied this commonet over to #110 and replied there: 
https://github.com/openshift/local-storage-operator/pull/110#discussion_r457153757

pkg/apis/local/v1alpha1/localvolumediscovery_types.go
```diff
@@ -0,0 +1,79 @@
...
+	NodeSelector *corev1.NodeSelector `json:"nodeSelector,omitempty"`
+	// If specified tolerations is the list of toleration that is passed to the
+	// LocalVolumeDiscovery Daemon
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+}
+
+// LocalVolumeDiscoveryStatus defines the observed state of LocalVolumeDiscovery
+type LocalVolumeDiscoveryStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
gnufied, line 40: nit: I do not think these placeholder comments belong in final API.

sp98: removed this comment along with a few other similar ones.

pkg/apis/local/v1alpha1/localvolumediscoveryresult_types.go
```diff
@@ -0,0 +1,95 @@
...
+	// Serial number of the disk
+	Serial string `json:"serial"`
+	// Size of the discovered device
+	Size resource.Quantity `json:"size"`
+	// Property represents whether the device type is rotational or not
+	Property DeviceMechanicalProperty `json:"property"`
+	// FSType represents the filesystem available on the device
+	FSType string `json:"fstype"`
+	// Status defines whether the device is available for use or not
+	Status DeviceStatus `json:"status"`
```
gnufied, line 47: this basically means everything is required here. But for many of these fields no information may be available. For example - `fsType` may be only available if there is a filesystem on the disk and in which case what happens?

pkg/diskmaker/discovery/monitor.go
```diff
@@ -0,0 +1,123 @@
...
+	defer close(c)
+	var udevFilter []string
+
+	// return any add or remove events, but none that match device mapper
+	// events. string matching is case-insensitive
+	events := make(chan string)
+
+	// regex to ignore the list of devices where udev events should not
+	// trigger continous discovery
+	discoverUdev := "(?i)dm-[0-9]+,(?i)rbd[0-9]+,(?i)nbd[0-9]+"
```
gnufied, line 26: Lets move this string to a constant. But also, why does this variable have to be initialized at all, if we are only creating it for the purpose of splitting into list of filters? Shouldn't we just initialize the filter list directly?

pkg/diskmaker/discovery/monitor.go
```diff
@@ -0,0 +1,123 @@
...
+	go rawUdevBlockMonitor(events,
+		[]string{"(?i)add", "(?i)remove"},
+		udevFilter)
+
+	for {
+		event, ok := <-events
+		if !ok {
+			return
+		}
+		timeout := time.NewTimer(period)
```
gnufied, line 40: So as I understand this - this code here tries to serialize udev events that were received in 5second interval and send only one event to discover goroutine, so as each udev event does not trigger a full device discovery?

pkg/controller/add_localvolumediscovery.go
```diff
@@ -0,0 +1,10 @@
+package controller
+
+import (
+	"github.com/openshift/local-storage-operator/pkg/controller/localvolumediscovery"
+)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	AddToManagerFuncs = append(AddToManagerFuncs, localvolumediscovery.Add)
+}
```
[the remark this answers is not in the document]
sp98: moved AddToManager functions in to `add_controllers.go` file.

pkg/diskmaker/discovery/discovery.go
```diff
@@ -0,0 +1,247 @@
...
+func (discovery *DeviceDiscovery) Start() error {
+	klog.Info("starting device discovery")
+	err := discovery.ensureDiscoveryResultCR()
+	if err != nil {
+		return errors.Wrapf(err, "failed to create LocalVolumeDiscoveryResult resource.")
+	}
+
+	err = discovery.discoverDevices()
+	if err != nil {
+		errors.Wrapf(err, "failed to start device discovery")
```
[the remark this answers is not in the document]
sp98: Added events.

pkg/diskmaker/discovery/spec.go
```diff
@@ -0,0 +1,120 @@
+package discovery
```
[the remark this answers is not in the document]
sp98: renamed the file to `sync_result.go`

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -0,0 +1,192 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Define a discovery daemonset object
+	discoveryDaemon := nodedaemon.GenerateDiskMakerDaemonset(request, instance.Spec.Tolerations, getOwnerRefs(instance), *instance.Spec.NodeSelector)
```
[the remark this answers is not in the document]
sp98: As discussed in the LSO meeting, we would be taking this up later on after the dev freeze.

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return err
+	}
+
+	// TODO(user): Modify this to be the types you create that are owned by the primary resource
+	// Watch for changes to secondary resource Pods and requeue the owner LocalVolumeDiscovery
+	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &localv1alpha1.LocalVolumeDiscovery{},
+	})
```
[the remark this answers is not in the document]
sp98: Removed this watch. `DaemonSet` gets the `LocalVolumeDiscovery` owner reference in the `reconciler` method.

pkg/diskmaker/discovery/discovery.go
```diff
@@ -0,0 +1,262 @@
...
+// getDiscoverdDevices creates v1alpha1.DiscoveredDevice from internal.BlockDevices
+func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.DiscoveredDevice {
+	discoveredDevices := make([]v1alpha1.DiscoveredDevice, 0)
+	for _, blockDevice := range blockDevices {
+		deviceID, err := blockDevice.GetPathByID()
+		if err != nil {
+			klog.Warningf("failed to get persisent ID for the device %q. Error %v", blockDevice.Name, err)
+			deviceID = ""
+		}
+		size, err := resource.ParseQuantity(blockDevice.Size)
```
gnufied, line 172: One of the motivations of defining size field as `resource.Quantity` is, so as it can seen/used as "10Gi" or "100GiB" in api objects. Using these values as such (as integers) defeats the purpose.

pkg/diskmaker/discovery/discovery.go
```diff
@@ -0,0 +1,262 @@
...
+		e := diskmaker.NewSuccessEvent(diskmaker.UpdatedDiscoveredDeviceList, message, "")
+		discovery.eventSync.Report(e, discovery.localVolumeDiscovery)
+	}
+
+	return nil
+}
+
+// getValidBlockDevices fetchs all the block devices sutitable for discovery
+func getValidBlockDevices() ([]internal.BlockDevice, error) {
+	blockDevices, badRows, err := internal.ListBlockDevices()
```
gnufied, line 143: We are ignoring and not listing LVM devices. I think this might be a problem. Was there an agreement about how to display LVM devices? i.e - if there are logical volumes out of a volume-group then are we going to list individual logical volumes or just the devices that make up the volume-group?

sp98: We didn't have any discussions regarding the how to display LVM devices earlier. Also the deviceType in discovery is using same [deviceType](https://github.com/openshift/local-storage-operator/blob/2a671867f76c7b0d2e3ed0f89d1d55350b9b82f8/pkg/apis/local/v1alpha1/localvolumeset_types.go#L23) enum in auto provision which is restricted to options like `disk, part and loop`.

pkg/diskmaker/discovery/discovery.go
sp98: @gnufied Can we have this change as part of bug fix. Because this will cause breaking changes in the UI and UI code is already merged.

pkg/diskmaker/discovery/discovery.go
sp98: @gnufied Can we have this change as part of bug fix. Because this will cause breaking changes in the UI and UI code is already merged.

gnufied: The problem is `deviceType` guarantee only exists for auto-provisioning API. It has a explicit field for that. Auto-discovery has no such restrictions and is expected to report most devices.

pkg/diskmaker/discovery/discovery.go
gnufied: wait, what breaking changes in the UI? I do not understand

gnufied: what I am trying to say if we are going to report size as integer bytes, then there was no point in picking `resource.Quantity` type for the size field.

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -0,0 +1,219 @@
...
+			reqLogger.Error(err, "failed to update status")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, err
+	} else if opResult == controllerutil.OperationResultUpdated || opResult == controllerutil.OperationResultCreated {
+		reqLogger.Info("daemonset changed", "daemonset.Name", ds.GetName(), "op.Result", opResult)
+	}
+
+	instance.Status.Phase = localv1alpha1.Discovering
+	err = r.updateStatus(instance)
```
gnufied, line 103: We should set Conditions too.

sp98: Added conditions and updated the unit tests to test conditions.

gnufied: and `ObservedGeneration`

sp98: added `ObservedGeneration` in the status

pkg/diskmaker/discovery/discovery.go
rohantmp: UI calculates discovered capacity. It does not handle every available format of resource.Quantity yet.

PR #133, opened 2020-07-28
log eval symlink error if it happens

gnufied:
Throwing away errors is not a good idea. :(

### DIFF
--- a/pkg/diskmaker/diskmaker.go
+++ b/pkg/diskmaker/diskmaker.go
@@ -297,7 +297,7 @@ func (d *DiskMaker) findMatchingDisks(diskConfig *DiskConfig, deviceSet sets.Str
 func (d *DiskMaker) findDeviceByID(deviceID string) (string, string, error) {
 	diskDevPath, err := filepath.EvalSymlinks(deviceID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/diskmaker/diskmaker.go
```diff
@@ -297,7 +297,7 @@ func (d *DiskMaker) findMatchingDisks(diskConfig *DiskConfig, deviceSet sets.Str
 func (d *DiskMaker) findDeviceByID(deviceID string) (string, string, error) {
 	diskDevPath, err := filepath.EvalSymlinks(deviceID)
 	if err != nil {
-		return "", "", fmt.Errorf("unable to find device with id %s", deviceID)
+		return "", "", fmt.Errorf("unable to find device with id %s: %v", deviceID, err)
```
bertinatto, line 300: I'd suggest taking the chance and make the error message more generic: it assumes a "not found" error while EvalSymlinks may return something else

gnufied: I renamed it to: `unable to find device at path %s` , because technically it is still "not found" error but the message coming from `EvalSymLinks` should clarify to the reader what exactly happened.

PR #135, opened 2020-08-04
Bug 1862607:  [Auto discovery]: change device size format to int64

sp98:
~~Currently, the discovery results show device size as integer bytes. This PR changes the size format to Mi, Gi or Ti.~~

Update: 
This PR changes the device size format to from `resource.Quanity` to `int64` so that it can be consumed easily by the higher level tools, like the UI


pkg/diskmaker/discovery/discovery.go
```diff
@@ -260,3 +263,28 @@ func parseDeviceType(deviceType string) v1alpha1.DeviceType {
...
+	MiB = 1024 * 1024
+	GiB = MiB * 1024
+	TiB = GiB * 1024
+
+	// Round off is required because Quantity needs a whole number.
+	// Decimal values get converted into milibytes.
+	// For example, 5.5Gi is converted into 5632Mi
+	if floatBytes < GiB {
+		num = int64(math.Ceil(floatBytes / MiB))
+		suffix = "Mi"
```
sp98, line 280: @gnufied do we need all three - Mi, Gi and Ti ? or Only Gi and Ti ?

pkg/diskmaker/discovery/discovery.go
```diff
@@ -260,3 +263,28 @@ func parseDeviceType(deviceType string) v1alpha1.DeviceType {
...
+	var num int64
+	var MiB, GiB, TiB float64
+	var suffix string
+	MiB = 1024 * 1024
+	GiB = MiB * 1024
+	TiB = GiB * 1024
+
+	// Round off is required because Quantity needs a whole number.
+	// Decimal values get converted into milibytes.
+	// For example, 5.5Gi is converted into 5632Mi
```
sp98, line 277: @gnufied Any suggestions/opinions on this rounding off behavior? (see above comment)

gnufied:
@sp98 thinking some more since size field is intended to be consumed by higher level tools (such as GUI) , we could keep reporting values as bytes. But IMO if we are going to report size as bytes then we should move the field to int64, so as it does not undergo unnecessary conversion when treating it as `resource.Quanity`

sp98:
Looking into the CI failure

PR #136, opened 2020-08-06
Bug 1862610: discover lvm devices in LocalVolumeDiscovery

sp98:
Local volume discovery is only discovering `disk` and `partition` device type. This PR adds support for discovering `lvm` device types as well. 

### DIFF
--- a/pkg/diskmaker/discovery/discovery.go
+++ b/pkg/diskmaker/discovery/discovery.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/local-storage-operator/pkg/diskmaker/controllers/lvset"
 	"github.com/openshift/local-storage-operator/pkg/internal"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog"
 )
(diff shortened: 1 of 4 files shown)


pkg/apis/local/v1alpha1/localvolumeset_types.go
```diff
@@ -29,6 +29,8 @@ const (
 	Partition DeviceType = "part"
 	// Loop type device
 	Loop DeviceType = "loop"
+	// LVMType is an LVM type
+	LVMType DeviceType = "lvm"
```
gnufied, line 33: this also changes auto-provisioning implementation which we shouldn't. cc @rohantmp

sp98: using `DiscoveredDevicesType` type for LocalVolumeDiscovery so that I doesn't change the auto-provisioning implementation.

pkg/diskmaker/discovery/discovery.go
```diff
@@ -211,7 +211,7 @@ func ignoreDevices(dev internal.BlockDevice) bool {
 		return true
 	}
 
-	if !(dev.Type == "disk" || dev.Type == "part") {
+	if !(dev.Type == "disk" || dev.Type == "part" || dev.Type == "lvm") {
```
gnufied, line 214: it might be cleaner to make a `set.String` and check if `set.Has(dev.Type)`?

sp98: using `set.String`

PR #137, opened 2020-08-06
Bug 1756096: Included must-gather script and image

huffmanca:
This PR adds a script and corresponding image to work with `oc adm must-gather`. This will cycle through all of the CRDs and output any resources for them.

### DIFF
--- a/must-gather/gather
+++ b/must-gather/gather
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+resources=()
+
+for I in $(/usr/bin/oc get crd | grep local.storage.openshift.io | awk '{print $1}'); do
+  resources+=($I)
+done
+
+for resource in ${resources[@]}; do
+  /usr/bin/oc get ${resource} --all-namespaces -o custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --no-headers 2> /dev/null | \
+  while read ocresource; do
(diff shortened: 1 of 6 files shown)


must-gather/gather
```diff
@@ -0,0 +1,26 @@
...
+      /usr/bin/oc get ${resource} -o yaml ${ocobject} > ${object_collection_path}/${ocobject}.yaml
+    else
+      object_collection_path=must-gather/namespaces/${ocns}/crs/${resource}
+      mkdir -p ${object_collection_path}
+      /usr/bin/oc get ${resource} -n ${ocns} -o yaml ${ocobject} > ${object_collection_path}/${ocobject}.yaml
+    fi
+  done
+done
+
+exit 0
```
gnufied, line 26: Can you add a quick doc about how to use must-gather with LSO? I would assume we will need more detailed docs in openshift-docs

huffmanca: I've included docs in a separate commit, and referenced this from the main README. I've included the current image name, and will update it if this is modified.

gnufied: thank you. one more question - does this supplement image needs to be built and shipped by ART pipeline?

huffmanca: It won't automatically get picked up. We'll need to file JIRAs to get ART to build it.

PR #145, opened 2020-08-26
 BUG 1869497:   fix invalid ownerrefernce for localvolumediscoveryresult CR

sp98:
LocalVolumeDiscovery CR uses ownerrefernce>kind as LocalVolumeDisovery but uses the ownerrefernce UUID of daemonset POD

This PR changes the owner references of the LocalVolumeDiscoveryResult objects to LocalVolumeDiscovery object. The reconcile method deletes the orphan results on every run. 

### DIFF
--- a/pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
+++ b/pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	localv1alpha1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1alpha1"
 	"github.com/openshift/local-storage-operator/pkg/common"
@@ -13,8 +14,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	corev1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -44,7 +47,11 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileLocalVolumeDiscovery{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileLocalVolumeDiscovery{
+		client:    mgr.GetClient(),
+		scheme:    mgr.GetScheme(),
+		reqLogger: log,
+	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -77,8 +84,9 @@ var _ reconcile.Reconciler = &ReconcileLocalVolumeDiscovery{}
 type ReconcileLocalVolumeDiscovery struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
+	client    client.Client
+	scheme    *runtime.Scheme
+	reqLogger logr.Logger
 }
 
 // Reconcile reads that state of the cluster for a LocalVolumeDiscovery object and makes changes based on the state read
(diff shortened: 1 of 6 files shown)


pkg/diskmaker/discovery/sync_result.go
```diff
@@ -30,7 +30,7 @@ func newDiscoveryResultInstance(nodeName, namespace, parentObjName, parentObjUID
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					APIVersion: v1alpha1.SchemeGroupVersion.String(),
-					Kind:       "LocalVolumeDiscovery",
```
gnufied, line 33: I do not understand why should discovery results object have owneref to pod or daemonset. Shouldn't this be `LocalVolumeDiscovery` object and its UUID.

sp98: I might be wrong, but how about the following scenario:
1-  User startts Discovery on Node1 and Node2.
2- User updates the Discovery CR to run it only on Node1. 

After Step 2,  daemonset pod running on `Node2` is gone so corresponding discovery result CR will be deleted. 
But if the Discovery Result CR was dependent on `LocalVolumeDiscovery`, then after step 2, daemonset pod will be deleted but the discovery result CR (on Node1) won't be deleted. 

So in short, if we have LocalVolumeDiscovery and its UUID as owner, the results will be deleted only when user deletes the LocalVolumeDiscovery CR complete and not when user updates it.

gnufied: I don't think it is a good idea to depend on ownereref mechanism for cleaning up leftover result objects. We already have established pattern of verifying if node selector includes certain node and may be we should periodically reconcile and delete leftover discovery result objects and use LocalVolumeDiscovery object as owneref.

sp98: Identifying `leftover discovery result objects` approach might get complicated if we decide to go with multiple `LocalVolumeDiscovery` objects in the future.  Right now, we mandate the user to create only a single `LocalVolumeDiscovery` object. 

With multiple `LocalVolumeDisovery` objects, a node might be removed from one object's nodeSelector, but it still might be part of another object's nodeSelector.

sp98: @gnufied Updated the owner reference.  LocalVolumeDiscoveryResults have owner reference of LocalVolumeDiscovery. Also deleting any orphan LocalVolumeDiscoveryResult on every reconcile.

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -283,13 +328,15 @@ func getOwnerRefs(cr *localv1alpha1.LocalVolumeDiscovery) []metav1.OwnerReferenc
...
-	return corev1.EnvVar{
-		Name: "UID",
-		ValueFrom: &corev1.EnvVarSource{
-			FieldRef: &corev1.ObjectFieldSelector{
-				FieldPath: "metadata.uid",
-			},
+func getEnvVars(objName, uid string) []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{
+			Name:  "UID",
```
gnufied, line 334: Can this environment variable be suitable named? `UID` as a generic name can conflict with other things.

pkg/diskmaker/discovery/sync_result.go
```diff
@@ -49,23 +49,14 @@ func (discovery *DeviceDiscovery) ensureDiscoveryResultCR() error {
 	nodeName := os.Getenv("MY_NODE_NAME")
 	namespace := os.Getenv("WATCH_NAMESPACE")
 	parentObjUID := os.Getenv("UID")
-	parentObjName := os.Getenv("POD_NAME")
+	parentObjName := os.Getenv("DISCOVERY_OBJECT_NAME")
 	if nodeName == "" || namespace == "" || parentObjUID == "" || parentObjName == "" {
```
gnufied, line 53: should we have a unit test for proper owner ref on discovery result objects?

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -202,6 +217,36 @@ func (r *ReconcileLocalVolumeDiscovery) updateDiscoveryStatus(instance *localv1a
...
+
+	for _, discoveryResult := range discoveryResultList.Items {
+		nodeName := discoveryResult.Spec.NodeName
+		node := &corev1.Node{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
+		if err != nil {
+			return fmt.Errorf("failed to get instance of node %q", nodeName)
+		}
+
+		matches := corev1helper.MatchNodeSelectorTerms(instance.Spec.NodeSelector.NodeSelectorTerms,
```
gnufied, line 235: this causes LSO to crash, if no nodeSelector was specified in CR..

gnufied: please ensure that we have a test case for this case.

sp98: Thanks. Added a check to skip when no NodeSelector is specified [here](https://github.com/openshift/local-storage-operator/blob/9c72cb708080ed6693dc090f52ed3282ca3fb6cc/pkg/controller/localvolumediscovery/localvolumediscovery_controller.go#L227).  Added unit test for this case here [https://github.com/openshift/local-storage-operator/blob/9c72cb708080ed6693dc090f52ed3282ca3fb6cc/pkg/controller/localvolumediscovery/localvolumediscovery_controller_test.go#L275](url)

pkg/diskmaker/discovery/sync_result.go
sp98: We do have some existing tests for that [here](https://github.com/openshift/local-storage-operator/blob/9c72cb708080ed6693dc090f52ed3282ca3fb6cc/pkg/diskmaker/discovery/sync_result_test.go#L144-L145) Will these suffice?

pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
sp98: Updated name to `DISCOVERY_OBJECT_UID`

PR #147, opened 2020-09-03
BUG 1872079: change device status to NotAvailable if block mode pv is provisioned on it.

sp98:
Continuous discovery changes the device status to NotAvailable when pv is provisioned on it.
In case of filesystem mode pv, the discovery checks the filesystem on the device and makes it NotAvailable.
For block mode pv, the discovery checks the bind mount of the device in /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/<pv-name>/<pod-uuid>
and makes the device as NotAvailable if a bind mount is present.

### DIFF
--- a/pkg/internal/diskutil.go
+++ b/pkg/internal/diskutil.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -17,6 +18,7 @@ var (
 	ExecCommand          = exec.Command
 	FilePathGlob         = filepath.Glob
 	FilePathEvalSymLinks = filepath.EvalSymlinks
+	mountFile            = "/proc/1/mountinfo"
 )
 
 var ExecResult string
@@ -118,6 +120,30 @@ func (b BlockDevice) HasChildren() (bool, error) {
 		}
 	}
 	return false, nil
+}
+
+// HasBindMounts checks for bind mounts and returns mount point for a device by parsing `proc/1/mountinfo`.
+// HostPID should be set to true inside the POD spec to get details of host's mount points inside `proc/1/mountinfo`.
+func (b BlockDevice) HasBindMounts() (bool, string, error) {
+	data, err := ioutil.ReadFile(mountFile)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to read file %s: %v", mountFile, err)
+	}
+
+	mountString := string(data)
+	for _, mountInfo := range strings.Split(mountString, "\n") {
+		if strings.Contains(mountInfo, b.Name) {
+			mountInfoList := strings.Split(mountInfo, " ")
+			if len(mountInfoList) >= 10 {
+				// device source is 4th field for bind mounts and 10th for regular mounts
+				if mountInfoList[3] == fmt.Sprintf("/%s", b.Name) || mountInfoList[9] == fmt.Sprintf("/dev/%s", b.Name) {
+					return true, mountInfoList[4], nil
+				}
+			}
+		}
+	}
+
+	return false, "", nil
 }
 
 // GetDevPath for block device (/dev/sdx)
@@ -329,7 +355,6 @@ func executeCmdWithCombinedOutput(cmd *exec.Cmd) (string, error) {
 	if err != nil {
 		return "", err
 	}
-
 	return strings.TrimSpace(string(output)), nil
 }
 
--- a/pkg/internal/diskutil_test.go
+++ b/pkg/internal/diskutil_test.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


pkg/internal/diskutil.go
```diff
@@ -120,6 +120,29 @@ func (b BlockDevice) HasChildren() (bool, error) {
 	return false, nil
 }
 
+// HasBindMounts checks for bind mounts in targetDirectory using `findmnt` command
+// Returns true if targetDirectory (target) and device name (source) are present in the ouput
+func (b BlockDevice) HasBindMounts(targetDir string) (bool, error) {
+	args := []string{"--raw"}
+	cmd := ExecCommand("findmnt", args...)
+	output, err := executeCmdWithCombinedOutput(cmd)
+	if err != nil {
```
gnufied, line 129: Should we use `findmnt` or `cat /proc/1/mountinfo` ? I personally like using `/proc/1/mountinfo` because that means we depend on one less utility inside the pod.

sp98: using `cat /proc/1/mountinfo` to get the bind mount details.

gnufied: Is there any reason to use `cat` and exec? can you not just open the file `/proc/1/mountinfo` and read it - rather than using exec interface?

But there are other problems here:
1. looking for a substring (such as `sda`) can match the mount path inadvertently.  according to - https://man7.org/linux/man-pages/man5/proc.5.html , we should only search for device name match in 4th and 10th field. 
2. If we correct device name matching as above, we don't need to know kubelet mount directory here. If a mount exists for the device then device can't be used.

sp98: > Is there any reason to use `cat` and exec? can you not just open the file `/proc/1/mountinfo` and read it - rather than using exec interface?
Updated code to read contents of the file. 

> 1. looking for a substring (such as `sda`) can match the mount path inadvertently.  according to - https://man7.org/linux/man-pages/man5/proc.5.html , we should only search for device name match in 4th and 10th field (search for `mountinfo` format description in the linked doc)

Thanks for the pointers. I could make use of the 4th field to get the `root` and 5th field to get the `mountpoint`.  
```5595 121 0:6 **/sdc** **/var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/local-pv-343bdd9/6d9d33ae-408e-4bac-81f7-c0bc347a9667** rw shared:23 - devtmpfs devtmpfs rw,seclabel,size=32180404k,nr_inodes=8045101,mode=755```

File: pkg/internal/diskutil.go [hunk 2]
```diff
@@ -120,6 +122,27 @@ func (b BlockDevice) HasChildren() (bool, error) {
...
+	data, err := ioutil.ReadFile(mountFile)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to read file %s: %v", mountFile, err)
+	}
+
+	mountString := string(data)
+	for _, mountInfo := range strings.Split(mountString, "\n") {
+		if strings.Contains(mountInfo, b.Name) {
+			mountInfoList := strings.Split(mountInfo, " ")
+			if len(mountInfoList) > 5 && mountInfoList[3] == fmt.Sprintf("/%s", b.Name) {
```
gnufied, line 137: For bind mounts the 4th field is the device source but for regular mounts the device source will be present in 10th field and hence we should check for match with either 4th or 10th field.

pkg/internal/diskutil.go
```diff
@@ -120,6 +122,30 @@ func (b BlockDevice) HasChildren() (bool, error) {
...
+		return false, "", fmt.Errorf("failed to read file %s: %v", mountFile, err)
+	}
+
+	mountString := string(data)
+	for _, mountInfo := range strings.Split(mountString, "\n") {
+		if strings.Contains(mountInfo, b.Name) {
+			mountInfoList := strings.Split(mountInfo, " ")
+			if len(mountInfoList) >= 10 {
+				// device source is 4th field for bind mounts and 10th for regular mounts
+				if mountInfoList[3] == fmt.Sprintf("/%s", b.Name) || mountInfoList[9] == fmt.Sprintf("/dev/%s", b.Name) {
```
sp98, line 139: I see that 4th field is usually `/<device-name>` and 10th field is `/dev/<device-name>`.  It is safe to assume that and use exact matches or should go with `string contains`?

pkg/internal/diskutil.go
sp98: Thanks. Updated.

gnufied:
oops I see that there is a follow up question. putting the PR on hold for a bit.

/hold

pkg/internal/diskutil.go
gnufied: as long as device name is real device name, I think it is okay to search for exact match.

pkg/internal/diskutil_test.go
```diff
@@ -185,6 +186,64 @@ func TestHasChildren(t *testing.T) {
...
+		},
+		{
+			label:              "Case 3", // device does not have mount points
+			blockDevice:        BlockDevice{Name: "sdd"},
+			mountInfo:          "5595 121 0:6 /sdc /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/local-pv-343bdd9/6d9d33ae-408e-4bac-81f7-c0bc347a9667 rw shared:23 - devtmpfs devtmpfs rw,seclabel,size=32180404k,nr_inodes=8045101,mode=755",
+			expected:           false,
+			expectedMountPoint: "",
+		},
+		{
+			label:              "Case 4", // device does not have mount points
```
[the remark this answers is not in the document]
sp98: No particular reason. Seen people using both ways. Just went with what I'm doing now. I can fix this in a separate PR.

PR #149, opened 2020-09-03
Bug 1874048: Prevent creation of duplicate PVs for same device

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1874048

cc @openshift/sig-storage

### DIFF
--- a/pkg/diskmaker/diskmaker.go
+++ b/pkg/diskmaker/diskmaker.go
@@ -8,10 +8,12 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/ghodss/yaml"
 	localv1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1"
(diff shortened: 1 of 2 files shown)


pkg/diskmaker/diskmaker.go
```diff
@@ -310,3 +341,95 @@ func fileExists(filename string) bool {
...
 }
+
+// getPVCreationLock checks whether a PV can be created based on this device
+// and Locks the device so that no PVs can be created on it while the lock is held.
+// the PV lock will fail if:
+// - another process holds an exclusive file lock on the device (using the syscall flock)
+// - a symlink to this device exists in symlinkDirs
+// returns:
+// ExclusiveFileLock, must be unlocked regardless of success
+// bool determines if flock was placed on device.
```
huffmanca, line 352: I know we return a bool indirectly via the exclusiveFileLock, but I wonder if this was part of a prior solution? Perhaps the comment here should be updated?

PR #150, opened 2020-09-09
Bug 1862429: Ensure that LocalVolumeSet cannot be deleted with released/bound PVCs.

rohantmp:


pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,41 @@
...
+		if f == finalizer {
+			return true
+		}
+	}
+	return false
+}
+
+// SetFinalizer either adds or removes finalizer depending on `setFinalizer`
+// it returns a bool indicating whether the metadata was changed or not.
+func SetFinalizer(metadata *metav1.ObjectMeta, finalizer string, setFinalizer bool) bool {
```
gnufied, line 17: Can we have two different functions for this  please?

pkg/controller/localvolumeset/finalizer.go
```diff
@@ -0,0 +1,60 @@
...
+	changed := false
+
+	if lvSet.DeletionTimestamp.IsZero() {
+		// finalizer should exist and be removed only when deleting
+		setFinalizer = true
+
+	} else {
+		// fetch PVs that match the storageclass
+		pvList := &corev1.PersistentVolumeList{}
+		err = r.client.List(context.TODO(), pvList, client.MatchingFields{pvStorageClassField: lvSet.Spec.StorageClassName})
```
gnufied, line 36: Using storageclass name in PV is not a good way of tracking PVs that were created by LSO.  storageclass can be deleted or can exist before localvolumset object was created. Or how about - I create 2 localvolumset objects with same SC name, now I can't delete one of them if other one is in use..

pkg/controller/localvolumeset/finalizer.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/openshift/local-storage-operator/pkg/common"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func (r *LocalVolumeSetReconciler) syncFinalizer(request reconcile.Request) error {
+	lvSet := &localv1alpha1.LocalVolumeSet{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, lvSet)
```
gnufied, line 17: Doesn't this function gets called after fetching localvolumeset object? Do we need to fetch it again?

pkg/common/finalizer_util.go
rohantmp: checking

pkg/controller/localvolumeset/finalizer.go
rohantmp: > I create 2 localvolumset objects with same SC name, now I can't delete one of them if other one is in use..

Since I don't think it'll be possible to associate a PV with a localvolumeset beyond the one-to-many mapping of the storagecluster relationship, I believe we will have to make sure there are no bound PVs associated with a storageclass before the corresponding lvsets are allowed to be removed as you say.

rohantmp: I guess this depends on the provisioner-as-a-library change

pkg/controller/localvolumeset/finalizer.go
rohantmp: The client reads from a cache, so it's not making a separate API call

pkg/controller/localvolumeset/finalizer.go
gnufied: It is possible to specify PV labels in configmap - https://github.com/openshift/local-storage-operator/blob/master/pkg/controller/localvolume/localvolume_controller.go#L803 which we are doing for localvolume objects. But in this case it is tricky because we are creating one configmap for all localvolumeset objects. I would settle for a label that tracks all of these PVs that they were created by `localvolumeset` object rather than waiting for provisioner-as-library change.

rohantmp: To identify objects created by a particular `localvolumeset` object or any `localvolumeset`? The latter can be done easily, but the former can't be done without a different configmap/provisioner-daemonset per `localvolumeset`

pkg/controller/localvolumeset/finalizer.go
gnufied: Where is this cache you speak of?

pkg/controller/localvolumeset/finalizer.go
gnufied: hmm yeah, there is no good way to track PVs created for particular `localvolumeset` object. This is however a problem - and I don't like to document yet another workaround.

rohantmp:
Aiming to rebase on #181 (allows associating PV with one LocalVolumeSet) by the end of the week.

pkg/common/finalizer_util.go
rohantmp: Hey, can you explain why you want to spit this? I'd have to rework the other the unit tests too, and It does seem cleaner to do:

```
if (something){
   needsFinalizer = false
else if (something else) { 
   needsFinalizer = true
}

err := setFinalizer(needsFinalizer)
// handle err

```

than

```
if (something){
   err := unSetFinalizer()
   //handle err
else if (something else) { 
   err := SetFinalizer()
  // handle err
}
```

pkg/controller/localvolumeset/finalizer.go
rohantmp: Have updated based on new LocalVolumeSet ownership labels

pkg/controller/localvolumeset/finalizer.go
rohantmp: [Here](https://github.com/kubernetes-sigs/controller-runtime/blob/dea6dee6c3e1ae2a2c0555f37ab28266105a1206/pkg/client/split.go#L115), but I'll change it because of uncertainty about when the cache is invalidated and there's not need to check if it's cached as we already have the value.

rohantmp:
will re open soon, have updated fully, just fixing newly added e2e

rohantmp:
@jsafrane @gnufied rebased and added e2e.

rohantmp:
Forgot to update unit tests to handle released pvs, oops.

rohantmp:
Updating to block deletion on released PVs too (since cleanup needs to happen). Leaving it convenient to modify so we can handle other reclaimPolicies in the future

rohantmp:
Updated lv and lvset to wait for both released and bound PVs.
@gnufied @jsafrane PTAL :)

rohantmp:
ugh, looks like the new e2e sent the time above the allowable limit:
```
Process did not finish before 25m0s timeout
```
This is why it took me so long to fix this e2e :disappointed: I'll try to make it shorter

Dockerfile.diskmaker
```diff
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/local-storage-operator
 COPY . .
 RUN make build-diskmaker
 
-FROM registry.ci.openshift.org/ocp/4.8:base
+FROM centos:8
```
priyanka19-98, line 6: Do we need this change?

rohantmp: Ah, good catch

rohantmp:
same thing, made it even shorter by 8 minutes with this last push

rohantmp:
strange, when I time it locally, it's at 16m and the limit says 25,I wonder if it's an underpowered container doing the build and it takes longer? Anyway, have one more thing to try before I have to further restructure tests.

rohantmp:
e2e succeeded this time, just adding one more tweak(making aws disk attaching async) because I've already done it :laughing:

jsafrane (commented):
did a quick check of the code, not the tests

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+		metadata.Finalizers = append(metadata.Finalizers, finalizer)
+	} else {
+		return false // not changed
+	}
+
+	return true
+}
+
+// GetBoundAndReleasedPVs owned by the object
+func GetBoundAndReleasedPVs(obj runtime.Object, c client.Client) ([]corev1.PersistentVolume, []corev1.PersistentVolume, error) {
```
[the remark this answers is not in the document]
rohantmp: Future proofing for when we support multiple reclaim policies.

pkg/controller/localvolume/reconcile.go
```diff
@@ -159,20 +159,23 @@ func (r *ReconcileLocalVolume) addSuccessCondition(lv *localv1.LocalVolume) *loc
...
+	pendingPVs := append(boundPVs, releasedPVs...)
+	if len(pendingPVs) > 0 {
+		pvNames := ""
+		for _, pv := range pendingPVs {
+			pvNames += fmt.Sprintf(" %v", pv.Name)
 		}
-	}
-	if len(boundPVs) > 0 {
-		msg := fmt.Sprintf("localvolume %s has bound persistentvolumes in use", commontypes.LocalVolumeKey(lv))
+		klog.Infof("bound/released PVs found, not removing finalizer %q", pvNames)
```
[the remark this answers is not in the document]
rohantmp: the reconciler has a built-in rate-limiter when errors are returned

rohantmp:
> It is more explicit. A function named SetFinalizer should not remove the finalizer. The code above is not a good coding example

looking for k8s util libs to switch to

pkg/common/names.go
```diff
@@ -34,6 +34,10 @@ const (
 	// DiscoveryNodeLabelKey is the label key on the discovery result CR used to identify the node it belongs to.
 	// the value is the node's name
 	DiscoveryNodeLabel = "discovery-result-node"
+
+	// LocalVolumeProtectionFinalizer is set to ensure the provisioner daemonset and owning object stick around long
+	// enough to handle the PV reclaim policy.
+	LocalVolumeProtectionFinalizer = "storage.openshift.com/local-volume-protection"
```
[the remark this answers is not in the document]
rohantmp: This is the same finalizer that `LocalVolumes` already use. Just forgot to switch that to refer to this.

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
+
```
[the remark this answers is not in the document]
rohantmp: will remove the extra lines

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+		return []corev1.PersistentVolume{}, []corev1.PersistentVolume{}, fmt.Errorf("could not get object metadata accessor from obj: %+v", obj)
```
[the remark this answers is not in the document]
rohantmp: +1

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+		return []corev1.PersistentVolume{}, []corev1.PersistentVolume{}, fmt.Errorf("failed to list persistent volumes: %w", err)
+	}
+	var boundPVs = make([]corev1.PersistentVolume, 0, len(pvList.Items))
+	var releasedPVs = make([]corev1.PersistentVolume, 0, len(pvList.Items))
+
+	for _, pv := range pvList.Items {
+		// ensure ownership
+		ownerKind, found := pv.Labels[PVOwnerKindLabel]
+		if !found {
+			return boundPVs, releasedPVs, fmt.Errorf("PV owner info not found in PV.Labels: %+v", pv.Labels)
```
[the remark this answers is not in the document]
rohantmp: removing

rohantmp:
@jsafrane done! PTAL

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if kindMatches && nameMatches && namespaceMatches {
+			if pv.Status.Phase == corev1.VolumeBound {
+				boundPVs = append(boundPVs, pv)
+
+			} else if pv.Status.Phase == corev1.VolumeReleased {
+				releasedPVs = append(releasedPVs, pv)
+			}
+		}
+
```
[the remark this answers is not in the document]
rohantmp: done

pkg/common/finalizer_util.go
```diff
@@ -0,0 +1,112 @@
...
+		ownerNamespace, found := pv.Labels[PVOwnerNamespaceLabel]
+		if !found {
+			return boundPVs, releasedPVs, fmt.Errorf("PV owner info not found in PV.Labels: %+v", pv.Labels)
+		}
+		namespaceMatches := ownerNamespace == namespace
+
+		if kindMatches && nameMatches && namespaceMatches {
+			if pv.Status.Phase == corev1.VolumeBound {
+				boundPVs = append(boundPVs, pv)
+
```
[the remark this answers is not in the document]
rohantmp: done

pkg/common/finalizer_util.go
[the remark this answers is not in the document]
rohantmp: done

pkg/common/finalizer_util.go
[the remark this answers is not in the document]
rohantmp: done

pkg/controller/localvolume/reconcile.go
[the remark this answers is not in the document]
rohantmp:
```
// DefaultControllerRateLimiter is a no-arg constructor for a default rate limiter for a workqueue.  It has
// both overall and per-item rate limiting.  The overall is a token bucket and the per-item is exponential
func DefaultControllerRateLimiter() RateLimiter {
	return NewMaxOfRateLimiter(
		NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
		// 10 qps, 100 bucket size.  This is only for retry speed and its only the overall factor (not per item)
		&BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
	)
}
```

pkg/common/finalizer_util.go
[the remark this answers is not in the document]
rohantmp: ah, missed this one, fixed.

rohantmp:
fixed a spacing thing I missed and retriggered because prow flaked

pkg/controller/localvolume/reconcile.go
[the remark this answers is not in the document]
rohantmp: hmm, just saw the limit comment, will add

test/e2e/gomega_util.go
```diff
@@ -59,3 +105,178 @@ func eventuallyFindPVs(t *testing.T, f *framework.Framework, storageClassName st
...
+
+func consumePV(t *testing.T, ctx *framework.Context, pv corev1.PersistentVolume) (*corev1.PersistentVolumeClaim, *batchv1.Job, *corev1.Pod) {
+	t.Logf("consuming PV: %q", pv.Name)
+	matcher := gomega.NewWithT(t)
+	f := framework.Global
+	name := fmt.Sprintf("%s-consumer", pv.ObjectMeta.Name)
+	pvc := &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: f.WatchNamespace,
```
priyanka19-98, line 145: Its changed to `f.OperatorNamespace` in the latest test-frameork.

rohantmp: Ah thanks! Was just looking for the right one. When I checked the framework code, I thought I saw WatchNamespace getting populated, but looks like it's blank

rohantmp:
@jsafrane

rohantmp:
rebased

controllers/localvolumeset/localvolumeset_controller.go
```diff
@@ -87,7 +88,13 @@ func (r *LocalVolumeSetReconciler) reconcile(ctx context.Context, request reconc
 	}
 
 	// store a one to many association from storageClass to LocalVolumeSet
-	r.LvSetMap.DeregisterStorageClassOwner(lvSet.Spec.StorageClassName, request.NamespacedName)
+	r.LvSetMap.RegisterStorageClassOwner(lvSet.Spec.StorageClassName, request.NamespacedName)
+
```
gnufied, line 92: is this a new bug?

controllers/localvolumeset/finalizer.go
```diff
@@ -0,0 +1,60 @@
...
+				// only print up to 10 PV names
+				if i >= 9 {
+					pvNames += "..."
+					break
+				}
+			}
+			r.ReqLogger.Info("bound/released PVs found, not removing finalizer", "pvNames", pvNames)
+		}
+	}
+
```
gnufied, line 46: there seems to be fair bit of duplication between this function and `cleanupLocalVolumeDeployment` - we should clean that up.

controllers/localvolumeset/localvolumeset_controller.go
```diff
@@ -87,7 +88,13 @@ func (r *LocalVolumeSetReconciler) reconcile(ctx context.Context, request reconc
...
+	// handle the LocalVolumeSet finalizer
+	err = r.syncFinalizer(*lvSet)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to update localvolumeset finalizer: %w", err)
+	}
```
gnufied, line 97: iirc - storageclass can't be deleted via ownerreferences and hence must be deleted manually?

rohantmp: Currently, it has to be deleted manually. We can have it deleted by finalizer in a following PR?

controllers/localvolumeset/finalizer.go
rohantmp: yes, in a follow-up?

controllers/localvolumeset/localvolumeset_controller.go
rohantmp: I think this bug happened with the migration PR

rohantmp:
rebased again

PR #156, opened 2020-09-25
Bug 1873424: Bios boot device is provisioned when devicetype is part

rohantmp:


pkg/internal/diskutil.go
```diff
@@ -28,6 +28,10 @@ const (
 	StateSuspended = "suspended"
 	// DiskByIDDir is the path for symlinks to the device by id.
 	DiskByIDDir = "/dev/disk/by-id/"
+	// BIOSBoot is the partition label found on bios partitions
+	BIOSBoot = "BIOS-BOOT"
+	// EFISystem is the partition label found on EFI partitions
+	EFISystem = "EFI-SYSTEM"
 )
```
gnufied, line 35: Are these strings fixed? My my desktop I see following label for efi partition - 

```
NAME="nvme0n1p1" ROTA="0" TYPE="part" SIZE="629145600" MODEL="" VENDOR="" RO="0" RM="0" STATE="" FSTYPE="vfat" KNAME="nvme0n1p1" SERIAL="" PARTLABEL="EFI System Partition"
```

I don't have a bios partition to test things but can we not use fstype present on these disks rather than partlabels?

rohantmp: Can we do a regex match (case insensitive and substring matching) against `bios.boot` and `efi.system`?

Can't think of anything better, @sp98 any ideas?

sp98: looks like we would need some regex here.

gnufied: for efi file system at least, we can use `vfat` filesystem rather than part label. I am not sure why we need to check for partition label for EFI disks. They are known to have FAT filesystem on them. Shouldn't they be automatically excluded?

BIOS or old MBR disks are another story - I don't even have one to verify. :(

rohantmp: fixed as per discussion

PR #157, opened 2020-09-25
Bug 1862120: LocalVolumeSet should delay provisioning after discovery so it won't race with other provisioners.

rohantmp:


pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,31 @@
...
+	deviceMinAge = time.Minute
+)
+
+func getTime() time.Time {
+	return time.Now()
+}
+
+// getFirstObserved returns the age that the key was first observed and records it if
+// it was first observed now
+func getFirstObserved(deviceAgeMap *map[string]time.Time, mux *sync.Mutex, key string, getTime func() time.Time) time.Time {
```
gnufied, line 20: I think this might be better written as a new type with mutex and map as properties of the new type. Something like:

```
type DeviceObservationAge struct{
  deviceAgeMap map[string]string
  mu *sync.Mutex
}
```

rohantmp: done

rohantmp:
verified after changes: 
```
Events:
  Type    Reason               Age                From                               Message
  ----    ------               ----               ----                               -------
  Normal  DiscoveredNewDevice  62s (x6 over 62s)  localvolumeset-symlink-controller  minikube - found possible matching disk, waiting 1m0s to claim
  Normal  FoundMatchingDisk    2s (x3 over 2s)    localvolumeset-symlink-controller  minikube - symlinking matching disk
```

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,52 @@
...
+var (
+	// deviceMinAge is the minimum age for a device to be considered safe to claim
+	// otherwise, it could be a device that some other entity has attached and we have not claimed.
+	deviceMinAge = time.Minute
+)
+
+type ageMap struct {
+	ageMap  map[string]time.Time
+	mux     sync.Mutex
+	getTime *func() time.Time
```
gnufied, line 17: Can we rename this function to `currentTime` which is what actually is. Also add a comment here that this is just for test purpose?

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -136,20 +136,33 @@ func (r *ReconcileLocalVolumeSet) Reconcile(request reconcile.Request) (reconcil
...
-) []internal.BlockDevice {
+) ([]internal.BlockDevice, []internal.BlockDevice) {
 	validDevices := make([]internal.BlockDevice, 0)
+	delayedDevices := make([]internal.BlockDevice, 0)
 	// get valid devices
 DeviceLoop:
 	for _, blockDevice := range blockDevices {
 
+		// get/set firstObserved
+		firstObserved := r.deviceAgeMap.getFirstObserved(blockDevice.KName)
```
gnufied, line 164: it might be cleaner to rename this function as  - `availableAt` so as this function reports time at which device will become available here.

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -165,6 +178,26 @@ DeviceLoop:
 				continue DeviceLoop
 			}
 		}
+
+		getTime := *r.deviceAgeMap.getTime
+
+		// skip devices younger than deviceMinAge, but remember them since they pass filters
+		if getTime().Sub(firstObserved) < deviceMinAge {
```
gnufied, line 185: if we rename function above then, this could be written as  - `if currentTime.After(availableAt) {}` .

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: hmm, I tried to keep `ageMap` generic. this would compromise that

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -165,6 +178,26 @@ DeviceLoop:
 				continue DeviceLoop
 			}
 		}
+
+		getTime := *r.deviceAgeMap.getTime
```
gnufied, line 182: I am not sure about - a function that just returns `time.Now()` leaking through here.

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: how about this? also removes `getTime`:
```
// get/set firstObserved
		isOldEnough := r.deviceAgeMap.isOlderThan(blockDevice.KName, deviceMinAge)

// 		// skip devices younger than deviceMinAge, but remember them since they pass filters
		if isOldEnough {
```

rohantmp:
addressed reviews, everything still works:
```
Events:
  Type    Reason               Age              From                               Message
  ----    ------               ----             ----                               -------
  Normal  DiscoveredNewDevice  4m (x6 over 4m)  localvolumeset-symlink-controller  minikube - found possible matching disk, waiting 1m0s to claim
  Normal  FoundMatchingDisk    3m (x3 over 3m)  localvolumeset-symlink-controller  minikube - symlinking matching disk
```

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -165,6 +179,24 @@ DeviceLoop:
 				continue DeviceLoop
 			}
 		}
+
+		// skip devices younger than deviceMinAge
+		if isOldEnough {
```
gnufied, line 184: but now this function and variable reads weird. It says - `if isOldEnough` which actually signals the opposite that the device is actually younger! can the function and variable be named appropriately?

rohantmp: not sure I follow: `isOldEnough` signals that the device is older than `minDeviceAge`

rohantmp: is it the comment?

rohantmp: I get it, I'll do 
```
if !isOldEnough {
   continue DeviceLoop
}
```

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,62 @@
...
+	f := func() time.Time {
+		return time.Now()
+	}
+	return &ageMap{currentTime: &f}
+}
+
+// checks if older than,
+// records current time if this is the first observation of key
+func (a *ageMap) isOlderThan(key string, t time.Duration) bool {
+
```
gnufied, line 30: unnecessary new line.

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,62 @@
...
+	firstObserved := a.getFirstObserved(key)
+	currentTime := *a.currentTime
+	return currentTime().Sub(firstObserved) > t
+
```
gnufied, line 34: unnecessary new line.

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// getFirstObserved returns the age that the key was first observed and records it if
+// it was first observed now
+func (a *ageMap) getFirstObserved(key string) time.Time {
```
gnufied, line 39: `getFirstObserved` is not merely getting the value, it is also storing and updating the map. IMO - a better name is needed for this function because a casual reader of code may think that this function is readonly..

pkg/diskmaker/controllers/lvset/device_age_test.go
```diff
@@ -0,0 +1,100 @@
...
+	for run, atTime := range runs {
+		t.Logf("Run %d, time is set to %+v", run, atTime)
+		// freeze time
+		r.deviceAgeMap.freezeTimeAt(atTime)
+		t.Logf("Adding %v devices at this time ^", increment)
+		// initial block devices (Set 0)
+		targetLength := len(blockDevices) + increment
+		for i := len(blockDevices); i < targetLength; i++ {
+			blockDevices = append(blockDevices, internal.BlockDevice{KName: fmt.Sprintf("dev-%d", len(blockDevices))})
+
```
gnufied, line 75: unnecessary new line.

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,62 @@
...
+type ageMap struct {
+	ageMap      map[string]time.Time
+	mux         sync.Mutex
+	currentTime *func() time.Time
+}
+
+func newAgeMap() *ageMap {
+	f := func() time.Time {
+		return time.Now()
+	}
```
gnufied, line 23: I think using a function pointer to set time for test purposes looks kind of weird. Not right now but may be we should consider defining mock interface for time and then using it.

pkg/diskmaker/controllers/lvset/device_age.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+var (
+	// deviceMinAge is the minimum age for a device to be considered safe to claim
+	// otherwise, it could be a device that some other entity has attached and we have not claimed.
+	deviceMinAge = time.Minute
```
gnufied, line 11: we can do this in a follow up but IMO this should be configurable via CLI option in diskmaker, so as certain deployments can run with different values potentially without requiring code change. This however may require propagating upto `LocalVolumeSet` object if this has to be configurable. We can do this in a follow up.

PR #158, opened 2020-09-25
Bug 1882775: Prevent creation of duplicate PVs for same device

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1882775

### DIFF
--- a/pkg/diskmaker/diskmaker.go
+++ b/pkg/diskmaker/diskmaker.go
@@ -8,10 +8,12 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/ghodss/yaml"
 	localv1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1"
+	"golang.org/x/sys/unix"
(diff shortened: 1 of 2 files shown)


pkg/diskmaker/diskmaker.go
```diff
@@ -310,3 +341,95 @@ func fileExists(filename string) bool {
 	}
 	return true
 }
+
+// getPVCreationLock checks whether a PV can be created based on this device
+// and Locks the device so that no PVs can be created on it while the lock is held.
+// the PV lock will fail if:
+// - another process holds an exclusive file lock on the device (using the syscall flock)
+// - a symlink to this device exists in symlinkDirs
+// returns:
```
huffmanca, line 350: This list of returns doesn't seem to be correct any longer.

huffmanca (commented):
Just one nitpick involving comments. I've reviewed the logic, and this makes sense to me.

pkg/diskmaker/diskmaker.go
gnufied: fixed

PR #159, opened 2020-09-25
Bug 1882630: lvset - configure default block cleaner command

rohantmp:
This will clean up the PV when it is released/deleted

### DIFF
--- a/pkg/controller/nodedaemon/provisioner_config.go
+++ b/pkg/controller/nodedaemon/provisioner_config.go
@@ -39,10 +39,11 @@ func (r *DaemonReconciler) reconcileProvisionerConfigMap(
 		storageClassName := lvSet.Spec.StorageClassName
 		symlinkDir := path.Join(common.GetLocalDiskLocationPath(), storageClassName)
 		mountConfig := localStaticProvisioner.MountConfig{
-			FsType:     lvSet.Spec.FSType,
-			HostDir:    symlinkDir,
-			MountDir:   symlinkDir,
-			VolumeMode: string(lvSet.Spec.VolumeMode),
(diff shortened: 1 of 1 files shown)


pkg/controller/nodedaemon/provisioner_config.go
```diff
@@ -39,10 +39,11 @@ func (r *DaemonReconciler) reconcileProvisionerConfigMap(
 		storageClassName := lvSet.Spec.StorageClassName
 		symlinkDir := path.Join(common.GetLocalDiskLocationPath(), storageClassName)
 		mountConfig := localStaticProvisioner.MountConfig{
-			FsType:     lvSet.Spec.FSType,
-			HostDir:    symlinkDir,
-			MountDir:   symlinkDir,
-			VolumeMode: string(lvSet.Spec.VolumeMode),
+			BlockCleanerCommand: []string{localStaticProvisioner.DefaultBlockCleanerCommand},
```
gnufied, line 42: After our discussion in google meet - I realized that if no `BlockCleanerCommand` is specified then static-provisioner seem to default to one:

```
if config.BlockCleanerCommand == nil {
			// Supply a default block cleaner command.
			config.BlockCleanerCommand = []string{DefaultBlockCleanerCommand}
		} else {
			// Validate that array is non empty.
			if len(config.BlockCleanerCommand) < 1 {
				return fmt.Errorf("Invalid empty block cleaner command for class %v", class)
			}
		}
```

We don't specify one for `LocalVolume` objects as well. So are you sure that this is what is causing the aforementioned bug?

rohantmp: no, not sure, but I haven't been able to reproduce it after trying a lot. Will ask on the bug for help reproing

PR #165, opened 2020-10-01
Bug 1884285: Do not hardcode GOOS/GOARCH

yselkowitz:


Makefile
```diff
@@ -6,8 +6,8 @@ ifeq ($(VERSION),)
 	VERSION = latest
 endif
 
-TARGET_GOOS=linux
-TARGET_GOARCH=amd64
+TARGET_GOOS=$(shell go env GOOS)
+TARGET_GOARCH=$(shell go env GOARCH)
```
wking, line 10: Dunno if it matters, but for the cluster-version operator, @abhinavdahiya was also [involving `GOHOSTOS` and `GOHOSTARCH`][1].

[1]: https://github.com/openshift/cluster-version-operator/commit/ddeb140857b9c59216d0c55d8167d7e5a0bfe325#diff-9283775b0feecc10455ec28bd08983b0R9-R12

gnufied: I agree that seems to be better default.

PR #166, opened 2020-10-02
Bug 1884285: Do not link with -extldflags "-static"

yselkowitz:
This breaks the build on ppc64le, which depends on the external linker
and therefore behaves differently than other architectures.


jsafrane:
/approve
(didn't /lgtm mean approval?)

yselkowitz:
LGTM only implies approved from an approver (owner).

PR #174, opened 2020-10-21
Bug 1887046: Remove unnecessary events

bertinatto:
QE claimed these events may cause confusion if there is only one LocalVolume or LocalVolumeSet.

https://bugzilla.redhat.com/show_bug.cgi?id=1887046


bertinatto:
CC @openshift/storage @sp98 @rohantmp

PR #175, opened 2020-10-22
Document testing using opm command

gnufied:
Document how to use new index image format.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -7,24 +7,60 @@ and then update `deploy/operator.yaml` to point to your images and follow rest o
 
 ## For developing on OpenShift and OLM
 
-1. Same as plain k8s, make your code change and build local-diskmaker and local-storage-operator images. Push them to quay or docker.io.
-2. Next we will have to update CSV inside manifests directory to point to those images.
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -7,24 +7,50 @@ and then update `deploy/operator.yaml` to point to your images and follow rest o
...
+```
+
+If you are using podman then there is no need to specify container-tool option.
+
+8. Tag and push index image to quay.io (or a container registry of your choice). Make sure that images are publicly available.
+
+9. Create a catalogSource and subscribe to the source by applying `examples/olm/catalog-create-subscribe.yaml`.
+
+```
+~> oc create -f examples/olm/catalog-create-subscribe.yaml
```
jsafrane, line 50: nit: I'd appreciate if there was explicit `vim examples/olm/catalog-create-subscribe.yaml` to replace the index image name with "your image"

jsafrane:
Tested the new procedure, works for me.

HACKING.md
gnufied: done

PR #181, opened 2020-11-11
Integrate static provisioner with diskmaker

rohantmp:
Provisioner libs:
- runtimeconfig is generated and PVs are created similarly to the provisioner, they still point at the symlink, but we will control the spec.

- populator (which is the PV cache for deleter) is populated from the eventhandler for diskmaker/controllers/lvset.Add
deleter (is run sequentially with the diskmaker because they access common resources that are not guaranteed to be threadsafe)

- [x] finish unit tests
- [ ] add integration tests for deletion
- [x] rethink events


pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -82,10 +90,41 @@ func (r *ReconcileLocalVolumeSet) Reconcile(request reconcile.Request) (reconcil
...
+	r.runtimeConfig.NodeLabelsForPV = provisionerConfig.NodeLabelsForPV
+	r.runtimeConfig.UseAlphaAPI = provisionerConfig.UseAlphaAPI
+	r.runtimeConfig.UseJobForCleaning = provisionerConfig.UseJobForCleaning
+	r.runtimeConfig.MinResyncPeriod = provisionerConfig.MinResyncPeriod
+	r.runtimeConfig.UseNodeNameOnly = provisionerConfig.UseNodeNameOnly
+	r.runtimeConfig.Namespace = request.Namespace
+	r.runtimeConfig.LabelsForPV = provisionerConfig.LabelsForPV
+	r.runtimeConfig.SetPVOwnerRef = provisionerConfig.SetPVOwnerRef
+	r.runtimeConfig.Name = getProvisionedByValue(*r.runtimeConfig.Node)
+
```
gnufied, line 103: I think this is a big surface area to copy from static-provisioner and prone to cause problems if they change their interfaces or we will be stuck with version we currently use.

pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -119,24 +166,39 @@ func (r *ReconcileLocalVolumeSet) Reconcile(request reconcile.Request) (reconcil
...
-		devLogger.Info("symlinking")
-		r.eventReporter.Report(lvset, newDiskEvent(diskmaker.FoundMatchingDisk, "symlinking matching disk", blockDevice.KName, corev1.EventTypeNormal))
-		err = symLinkDisk(lvset, r.eventReporter, devLogger, blockDevice, symLinkDir)
+		// Retrieve list of mount points to iterate through discovered paths (aka files) below
+		mountPoints, err := r.runtimeConfig.Mounter.List()
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("error retrieving mountpoints: %w", err)
+		}
+		// Put mount points into set for faster checks below
+		type empty struct{}
```
gnufied, line 180: nothing uses this `empty` type afaict.

pkg/diskmaker/controllers/lvset/create_pv.go
```diff
@@ -0,0 +1,175 @@
...
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	provCommon "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
+)
+
+func (r *ReconcileLocalVolumeSet) createPV(
```
gnufied, line 19: It would be nice to have a version of this function in static-provisioner which can be used directly. To be clear  creating the PV and saving the PV in api-server are two separate concerns and since we preferably want to add labels etc, we just want a unsaved PV to be returned by this function.

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: It's mostly required configuration for the deleter.  It's all a struct representation of their configuration file format, so we should be guaranteed some stability. That being said, we can probably get rid of these:
```
	r.runtimeConfig.UseAlphaAPI = provisionerConfig.UseAlphaAPI
	r.runtimeConfig.NodeLabelsForPV = provisionerConfig.NodeLabelsForPV
	r.runtimeConfig.UseJobForCleaning = provisionerConfig.UseJobForCleaning
	r.runtimeConfig.MinResyncPeriod = provisionerConfig.MinResyncPeriod
```
and hardcode these:
```
	r.runtimeConfig.UseNodeNameOnly = false
	r.runtimeConfig.SetPVOwnerRef = true
```

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: yes, switched it to sets.String, will get rid of it.

pkg/diskmaker/controllers/lvset/create_pv.go
rohantmp: static-provisioner does have one which we are using inside this function (which refers to create of create/apply, maybe we should go with apply):
```
	localPVConfig := &provCommon.LocalPVConfig{
		Name:            pvName,
		HostPath:        symLinkPath,
		Capacity:        common.RoundDownCapacityPretty(capacityBytes), // d.VolUtil.GetBlockCapacityByte(filePath)
		StorageClass:    storageClass.GetName(),
		ReclaimPolicy:   reclaimPolicy,        // fetch from storageClass (created by LocalVolumeSet operator controller)
		ProvisionerName: r.runtimeConfig.Name, // populate in runtimeconfig earlier
		VolumeMode:      desiredVolumeMode,
		MountOptions:    storageClass.MountOptions, // d.getMountOptionsFromStorageClass(class)
		SetPVOwnerRef:   true,                      // set owner reference from node to PV
		OwnerReference: &metav1.OwnerReference{
			Kind:       r.runtimeConfig.Node.Kind,
			APIVersion: r.runtimeConfig.Node.APIVersion,
			Name:       r.runtimeConfig.Node.GetName(),
			UID:        r.runtimeConfig.Node.GetUID(),
		},
		NodeAffinity: nodeAffinity,
	}
	if desiredVolumeMode == corev1.PersistentVolumeFilesystem && obj.Spec.FSType != "" {
		localPVConfig.FsType = &obj.Spec.FSType
	}
	newPV := provCommon.CreateLocalPVSpec(localPVConfig)
```

pkg/diskmaker/controllers/lvset/reconcile.go
rohantmp: First time we're bitten by this, we can switch to not using runtimeconfig inside our own methods.

Dockerfile.diskmaker
```diff
@@ -3,10 +3,14 @@ WORKDIR /go/src/github.com/openshift/local-storage-operator
 COPY . .
 RUN make build-diskmaker
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
```
rohantmp, line 6: Accidental change. Was experimenting with what binaries are available in builder, but now we're copying the necessary binaries, will revert in new commit.

All changes new changes will be pushed as new commits and rebased before merging.

rohantmp: Actually, can't figure out how to get `e2fsprogs` for mkfs.ext2 onto `registry.svc.ci.openshift.org/openshift/origin-v4.0:base` it is on `registry.svc.ci.openshift.org/openshift/release:golang-1.13` though

gnufied: This is what we did in static-provisioner fork - https://github.com/openshift/sig-storage-local-static-provisioner/blob/master/Dockerfile.rhel7 . We should not try and copy those binaries but install them

deploy/scripts/quick_reset.sh
```diff
@@ -0,0 +1,40 @@
...
+# Validate that we got a valid block device to cleanup
+validateBlockDevice
+
+echo "Calling mkfs"
+ionice -c 3 mkfs -F $LOCAL_PV_BLKDEVICE
+
+echo "Calling wipefs"
+ionice -c 3 wipefs -a $LOCAL_PV_BLKDEVICE
+
+echo "Quick reset completed"
```
gnufied, line 40: Can we copy these from `vendor/xx/static-provisioner/ ` during building the image, rather than copying them here?

gnufied: Oh well. I missed the discussion on slack about vendoring process missing these scripts.

Dockerfile.diskmaker
rohantmp: Hey, I tried that, but it's asking for a subscription to be activated when I tried that in `Dockerfile.diskmaker` it'll probably work in `Dockerfile.diskmaker.rhel7` since that's the same image. I'm not clear about the purpose of the two Dockerfiles, is the rhel one what we actually use downstream?  Is it ok if we just use the builder image for `Dockerfile.diskmaker`?

rohantmp: (builder is ~1GiB while origin base is ~300MiB)

gnufied: `Dockerfile.diskmaker` is used by us the developers. We can pretty much use anything as base here. But `Dockerfile.diskmaker.rhel7` is used by ART team and should not require subscription when it is installed.

rohantmp: ack, will go with this approach then

pkg/diskmaker/controllers/lvset/create_pv.go
gnufied: Yes this is still bit ugly IMO. We can clean this up later. I am thinking we need to make some changes in static-provisioner, so as it can be easy to use as a library. Currently it does not look like the case.

rohantmp:
@gnufied Added device labels to in another commit on top of these commits. Can I cherry pick that into this PR?

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -67,13 +67,5 @@ func (r *DaemonReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 		r.reqLogger.Info("daemonset changed", "daemonset.Name", ds.GetName(), "op.Result", opResult)
 	}
 
-	localProvisionerDSMutateFn := getLocalProvisionerDSMutateFn(request, tolerations, ownerRefs, nodeSelector, configMapDataHash)
```
gnufied, line 70: So why are we doing this only for `LocalVolumeSet` objects and not for `LocalVolume` objects?

gnufied: Lets fix this in a follow up.

PR #182, opened 2020-11-11
Modify the diskmaker to use internal BlockDevices

kesavanvt:
**Description**
Modified diskmaker to use internal.blockdevice
https://issues.redhat.com/browse/STOR-466 

Signed-off-by: kesavan <kvellalo@redhat.com>


pkg/diskmaker/diskmaker.go
```diff
@@ -184,24 +182,21 @@ func (d *DiskMaker) createSymLink(deviceNameLocation DiskLocation, symLinkPath s
...
-		klog.Errorf(msg)
+		klog.Errorf(msg, "could not list block devices", "lsblk.BadRows", badRows)
 		return
+	} else if len(badRows) > 0 {
+		msg := fmt.Sprintf("error parsing rows: %+v", badRows)
+		e := NewEvent(ErrorRunningBlockList, msg, "")
+		d.eventSync.Report(e, d.localVolume)
+		klog.Errorf(msg, "could not parse all the lsblk rows", "lsblk.BadRows", badRows)
 	}
-	deviceSet := d.findNewDisks(out.String())
```
sp98, line 202: Is `findNewDisks` not used anymore?  If that's the case, then this method implementation can be removed. Also, removing this method would make all the other methods in `disktable.go` redundant. So they should be removed as well. 

Also, the existing functionality was filtering the disks by ignoring disks based on whether they had `mountpoints` or if they were `root` disks with `partitions`.  See the `filterUsableDisks` method.  I don't see these filters applied after the new change. Is that expected ?

sp98 (commented):
@kesavanvt

pkg/diskmaker/diskmaker.go
kesavanvt: Thanks Santosh for pointing it out, i agree that `disktable.go` will not be used anywhere when `findNewDisks` is removed and have modified accordingly. Also, made changes to preserve the existing functionality.

rohantmp:
Looks good! Might be better to wait and rebase on #186

rohantmp (approved):
LGTM!

rohantmp:
@gnufied  PTAL

PR #186, opened 2020-11-20
Migrate LV controller to controller-runtime and shift it to the unified diskmaker

priyanka19-98:


gnufied:
Can you fix e2e please?

priyanka19-98:
> Can you fix e2e please?

Fixed!

priyanka19-98:
@rohantmp @gnufied , could you please review the PR.

rohantmp:
Yes, reviewing today.

cmd/diskmaker/main.go
```diff
@@ -18,8 +18,8 @@ var managerCmd = &cobra.Command{
 }
 var lvDaemonCmd = &cobra.Command{
 	Use:   "lv-controller",
-	Short: "Used to start an instance of the controller for a specific LocalVolume CR",
-	RunE:  startDiskMaker,
+	Short: "Used to start the controller-runtime manager that owns the LocalVolume controller",
+	RunE:  startManager,
```
rohantmp, line 22: if it's using the same function as lv-manager, no need to add a new command, just change the command invoked in `generateDiskMakerDaemonSet `to `lv-manager`

rohantmp: However, this will also start a lvset controller on every diskmaker daemonset spawned by localvolume_controller. Maybe we should just include the localvolume nodeSelectors in `/home/tpjsm/go/src/github.com/openshift/local-storage-operator/pkg/controller/nodedaemon/reconcile.go`/`lvSets, tolerations, ownerRefs, nodeSelector, err := r.aggregateDeamonInfo(request)`

the tolerations, ownerRefs, and nodeSelector should incorporate info from all localvolumes like it currently does for localvolumesets

rohantmp (changes requested):

rohantmp:
Looks good overall, just want to change it the single daemonset instead of multiple because the controller manger runs both controllers.

pkg/diskmaker/controllers/lv/diskmaker.go
```diff
@@ -255,22 +208,22 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
...
 	for storageClass, deviceArray := range deviceMap {
 		for _, deviceNameLocation := range deviceArray {
-			symLinkDirPath := path.Join(d.symlinkLocation, storageClass)
-			d.createSymLink(deviceNameLocation, symLinkDirPath)
+			symLinkDirPath := path.Join(r.symlinkLocation, storageClass)
+			r.createSymLink(deviceNameLocation, symLinkDirPath)
 		}
 	}
 
+	return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
priyanka19-98: Why do we need to Requeue?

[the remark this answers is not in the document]
rohantmp: To check for new devices. The previous implementation also used a ticker to achieve this

cmd/diskmaker/main.go
priyanka19-98: @rohantmp , as the `LocalVolumeSet`  has the `ProvisionerConfigMap, DiskMaker DaemonSet` in `nodedaemon controller` , the `ProvisionerConfigMap, DiskMaker DaemonSet` for LocalVolume should also move to nodedaemon, so that there is a single daemonset for all `LocalVolume CRs`. This will also require some functional tests to modify. Do you want me to do this in the same PR, or open a different one?

rohantmp: No need to change the provisioner configmap! And leave LocalVolume to create one local-static-provisioner per LocalVolume. We can fix that in a follow up where I integrate static-provisioner to LocalVolume

priyanka19-98: Thanks! Will update the PR to only change to the single ` Diskmaker Daemonset ` for LocalVolume.

pkg/controller/nodedaemon/daemonsets.go
```diff
@@ -91,6 +93,43 @@ func getDiskMakerDSMutateFn(
 	}
 }
 
+func getLVDiskMakerDSMutateFn(
```
rohantmp, line 96: we don't need a separate daemonset for LocalVolumes

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -49,13 +49,27 @@ func (r *DaemonReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
+	lvs, lvTolerations, lvOwnerRefs, lvNodeSeletor, err := r.aggregateLVDaemonInfo(request)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if len(lvs.Items) < 1 {
```
rohantmp, line 56: Now the reconcile won't finish if there aren't both LocalVolumes or LocalVolumeSets.

It should be:

```
if len(lvs.Items) < 1 || len(lvSets.Items) <1 {
    return reconcile.Resut{}, err
}
```
@priyanka19-98 this is why your diskmaker pods aren't coming up

priyanka19-98: What will be the err in
```
if len(lvs.Items) < 1 || len(lvSets.Items) <1 {
    return reconcile.Resut{}, err
}
```
?

pkg/controller/nodedaemon/daemonsets.go
priyanka19-98: How would the `DaemonSet` have both `localVolume, LocalVolumeSet` info?

rohantmp:
/retitle Migrate LV controller to controller-runtime and shift it to the unified diskmaker

pkg/diskmaker/controllers/lv/diskmaker.go
```diff
@@ -255,22 +262,35 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
...
-			d.createSymLink(deviceNameLocation, symLinkDirPath)
+			symLinkDirPath := path.Join(r.symlinkLocation, storageClass)
+			r.createSymLink(deviceNameLocation, symLinkDirPath)
 		}
 	}
 
+	return reconcile.Result{Requeue: true, RequeueAfter: checkDuration}, nil
+}
+
+func nodeSelectorMatchesNodeLabels(node *corev1.Node, nodeSelector *corev1.NodeSelector) (bool, error) {
```
rohantmp, line 280: since this is in both lv and lvset diskmaker controllers, can we move this to the `common` pkg?

pkg/controller/localvolume/localvolume_controller.go
```diff
@@ -144,66 +141,51 @@ func (r *ReconcileLocalVolume) syncLocalVolumeProvider(instance *localv1.LocalVo
...
-
 	children := []operatorv1.GenerationStatus{}
+	localProvisionerDS := &appsv1.DaemonSet{}
+	key := types.NamespacedName{Name: nodedaemon.ProvisionerName, Namespace: o.ObjectMeta.Namespace}
+	err = r.client.Get(context.TODO(), key, localProvisionerDS)
 
-	provisionerDS, err := r.syncProvisionerDaemonset(o, rollOutDaemonSet)
 	if err != nil {
-		klog.Errorf("failed to create daemonset for provisioner %s: %v", o.Name, err)
+		klog.Errorf("failed to create daemonset for provisioner %v", err)
```
rohantmp, line 156: It is now "failed to get"

pkg/controller/nodedaemon/provisioner_config.go
```diff
@@ -16,13 +16,15 @@ import (
...
 	localStaticProvisioner "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 
+	v1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1"
 	localv1alpha1 "github.com/openshift/local-storage-operator/pkg/apis/local/v1alpha1"
 	"github.com/openshift/local-storage-operator/pkg/common"
 )
 
 func (r *DaemonReconciler) reconcileProvisionerConfigMap(
 	request reconcile.Request,
 	lvSets []localv1alpha1.LocalVolumeSet,
```
rohantmp, line 26: lets remove lvSets from here

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	// ProvisionerName is the name of the local-static-provisioner daemonset
-	ProvisionerName = "localvolumeset-local-provisioner"
+	ProvisionerName = "local-provisioner"
```
rohantmp, line 24: This value is used in the `cleanupOldDaemonsets()` to cleanup older deamonsets, can you preserve this as `oldProvisionerName` (unexported) and change the usages in that function to that value?

rohantmp (changes requested):

test/e2e/localvolume_test.go
```diff
@@ -86,8 +86,8 @@ func LocalVolumeTest(ctx *framework.Context, cleanupFuncs *[]cleanupFn) func(*te
 			},
 		)
 
-		provisionerDSName := localVolume.Name + "-local-provisioner"
-		diskMakerDSName := localVolume.Name + "-local-diskmaker"
+		provisionerDSName := "local-provisioner"
```
rohantmp, line 89: you can import these values from `nodedaemon` so we don't have to update in two places in the future (and after e2e fails once :) )

gnufied:
hmm - it seems like we are making incremental changes to "partially" move to controller-runtime but still using separate dasemonset for PV provisioning. Is that what we want? Because I thought we are going to move volume provisioning into diskmaker entirely (and use static-provisioner as a library).

rohantmp:
@gnufied ,yup, we're going to include the provisioner-as-a-lib change in this PR

PR #187, opened 2020-11-23
use /disk/by-id name as symlink name in LocalVolume

sp98:
If the device provided in the `devicePaths` list in the localVolume CR has `/dev/disk/by-id`, then use the `id` name to create symlink in `/mnt/local-storage/<storageClassName>` directory. 

for example: `/mnt/local-storage/belowsixty/ata-VBOX_HARDDISK_____VB3ec86ebf-ed3a5a93`

Test:

1. /dev/sdb and /dev/sdc had /disk/by-id. So symlinks created by id
   /dev/sdd had no id. So Symlink got created by disk name

```
/mnt/local-storage/belowsixty
$ ls
ata-VBOX_HARDDISK_____VB3ec86ebf-ed3a5a93  ata-VBOX_HARDDISK_____VBddf003e5-4cfb17c1  sdd
```

2. Triggering warning events when device has no id
```
$ kubectl get event | grep  localvolume/example
16s         Normal    FoundMatchingDisk         localvolume/example                           minikube - found matching disk /dev/sdc with id /dev/disk/by-id/ata-VBOX_HARDDISK_____VB09c666fa-a3385f88
16s         Warning   SymlinkedOnDeivceName     localvolume/example                           minikube - created symlink on device name /dev/sdd with no disk/by-id. device name might not persist on reboot
16s         Normal    FoundMatchingDisk         localvolume/example                           minikube - found matching disk /dev/sdb with id /dev/disk/by-id/ata-VBOX_HARDDISK_____VB3ec86ebf-ed3a5a93
```

### DIFF
--- a/pkg/diskmaker/diskmaker.go
+++ b/pkg/diskmaker/diskmaker.go
@@ -124,10 +124,22 @@ func (d *DiskMaker) Run(stop <-chan struct{}) {
 	}
 }
 
-func (d *DiskMaker) createSymLink(deviceNameLocation DiskLocation, symLinkPath string, symLinkDirPath string, baseDeviceName string) {
+func (d *DiskMaker) createSymLink(deviceNameLocation DiskLocation, symLinkDirPath string) {
+
+	var symLinkSource, symLinkTarget string
+	var isSymLinkedByDeviceName bool
+	if deviceNameLocation.diskID != "" {
+		symLinkSource = deviceNameLocation.diskID
+		symLinkTarget = getSymlinkTarget(deviceNameLocation.diskID, symLinkDirPath)
+	} else {
+		symLinkSource = deviceNameLocation.diskNamePath
+		symLinkTarget = getSymlinkTarget(deviceNameLocation.diskNamePath, symLinkDirPath)
+		isSymLinkedByDeviceName = true
+	}
+
 	// get PV creation lock which checks for existing symlinks to this device
 	pvLock, pvLocked, existingSymlinks, err := internal.GetPVCreationLock(
-		deviceNameLocation.diskNamePath,
+		symLinkSource,
 		d.symlinkLocation,
 	)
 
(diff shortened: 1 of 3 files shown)


pkg/diskmaker/diskmaker.go
```diff
@@ -253,15 +253,25 @@ func (d *DiskMaker) symLinkDisks(diskConfig *DiskConfig) {
 		for _, deviceNameLocation := range deviceArray {
 
 			symLinkDirPath := path.Join(d.symlinkLocation, storageClass)
+			var symLinkPath string
+			if deviceNameLocation.diskID != "" {
+				symLinkPath = getSymlinkPath(deviceNameLocation.diskID, symLinkDirPath)
+			} else {
+				symLinkPath = getSymlinkPath(deviceNameLocation.diskNamePath, symLinkDirPath)
+			}
 			baseDeviceName := filepath.Base(deviceNameLocation.diskNamePath)
```
gnufied, line 262: And we need to event if we are going to use device name for symbolic link?

gnufied: Also since we know of issues around using device name even when device id is available, I am thinking we should warn users when they use device name in `LocalVolume` CR.

sp98: > And we need to event if we are going to use device name for symbolic link?

Added warning event when symbolic link source is device name. 

> Also since we know of issues around using device name even when device id is available, I am thinking we should warn users when they use device name in `LocalVolume` CR.

should this be a warning in the docs?

pkg/diskmaker/diskmaker.go
```diff
@@ -124,10 +124,22 @@ func (d *DiskMaker) Run(stop <-chan struct{}) {
...
+	var isSymLinkedByDeviceName bool
+	if deviceNameLocation.diskID != "" {
+		symLinkSource = deviceNameLocation.diskID
+		symLinkTarget = getSymlinkTarget(deviceNameLocation.diskID, symLinkDirPath)
+	} else {
+		symLinkSource = deviceNameLocation.diskNamePath
+		symLinkTarget = getSymlinkTarget(deviceNameLocation.diskNamePath, symLinkDirPath)
+		isSymLinkedByDeviceName = true
+	}
+
```
gnufied, line 139: Can we add some tests for this code?

pkg/diskmaker/diskmaker.go
gnufied: yeah we should catch this warning in a doc I think.

pkg/diskmaker/diskmaker.go
sp98: added unit tests for `createSymLink` method.

PR #177, opened 2020-10-27
Bug 1886973: Hide the LocalVolumeDiscoveryResult CRD from the UI

huffmanca:
Adds the LocalVolumeDiscoveryResult CRD to the list of internal objects, preventing it from being displayed in the UI.

PR #192, opened 2020-12-04
Bug 1904577: Hide the LocalVolumeDiscoveryResult CRD from the UI

huffmanca:
This is a cherrypick of https://github.com/openshift/local-storage-operator/pull/177 for 4.6.

Adds the LocalVolumeDiscoveryResult CRD to the list of internal objects, preventing it from being displayed in the UI.


gnufied:
I think there is something busted in the build. We need to investigate. 

/hold

PR #198, opened 2021-01-13
Bug 1897050: localVolumeSet: keep default MinSize to 1Gi and ignore devices with `boot` substring in the partLabel

sp98:
LocalVolumeSet doesn't cover all the possible cases to ignore boot devices and hence create
PVs on them. This PR keeps default minSize to 1Gi so that boot devices can be ignored. User can override the minSize value from the CR. Also ignore device where partlabel contains substring "boot"

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1897050

### DIFF
--- a/pkg/diskmaker/controllers/lvset/matcher.go
+++ b/pkg/diskmaker/controllers/lvset/matcher.go
@@ -15,7 +15,7 @@ const (
 	notReadOnly           = "notReadOnly"
 	notRemovable          = "notRemovable"
 	notSuspended          = "notSuspended"
-	noBiosInPartLabel     = "noBiosInPartLabel"
+	noBiosBootInPartLabel = "noBiosBootInPartLabel"
 	noFilesystemSignature = "noFilesystemSignature"
 	// file access , can't mock test
 	noChildren = "noChildren"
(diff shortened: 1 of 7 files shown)


pkg/diskmaker/controllers/lvset/matcher.go
```diff
@@ -98,11 +100,12 @@ var matcherMap = map[string]func(internal.BlockDevice, *localv1alpha1.DeviceIncl
...
-		if spec.MinSize != nil {
-			// quantity greater than min: -1
-			// quantity equal to min: 0
-			greaterThanOrEqualToMin = spec.MinSize.Cmp(quantity) <= 0
+		if spec.MinSize == nil {
+			spec.MinSize = &defaultMinSize
 		}
+		// quantity greater than min: -1
+		// quantity equal to min: 0
+		greaterThanOrEqualToMin = spec.MinSize.Cmp(quantity) <= 0
```
gnufied, line 109: How about also updating the rules so as if a partition has "boot" in substring, the partition is not selected? If they want to force use of such a partition they can always use `LocalVolume` object.

sp98: added in `noBiosBootInPartLabel` rule above.

PR #199, opened 2021-01-13
Bug 1914125: updated localVolume CRD and CSV files examples to encourage using /dev/disk/by-id...

sp98:

### DIFF
--- a/manifests/4.7/local-volumes.crd.yaml
+++ b/manifests/4.7/local-volumes.crd.yaml
@@ -64,7 +64,8 @@ spec:
                         type: string
                       devicePaths:
                         description: 'A list of devices which would be chosen for local storage.
-                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
+                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"].
+                        It is strongly recommended to use the unique /dev/disk/by-id/... name if available, as these identifiers persist across node restarts.'
                         items:
                           type: string
                         type: array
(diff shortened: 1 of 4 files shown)


manifests/4.7/local-volumes.crd.yaml
```diff
@@ -64,7 +64,9 @@ spec:
                         type: string
                       devicePaths:
                         description: 'A list of devices which would be chosen for local storage.
-                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
+                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]
+                        Note: Use the unique /dev/disk/by-id/.. name, if available, to avoid issues
```
huffmanca, line 68: Really just a nitpick, but I don't think we should have a Note in the CRD. Perhaps something similar to:

> It is strongly recommended to use the unique /dev/disk/by-id/... name if available, as these identifiers persist across node restarts.

opm-bundle/manifests/local-volumes.crd.yaml
```diff
@@ -64,7 +64,9 @@ spec:
                         type: string
                       devicePaths:
                         description: 'A list of devices which would be chosen for local storage.
-                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]'
+                        For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]
+                        Note: Use the unique /dev/disk/by-id/.. name, if available, to avoid issues
```
huffmanca, line 68: Same comment as above.

manifests/4.7/local-volumes.crd.yaml
sp98: updated.

opm-bundle/manifests/local-volumes.crd.yaml
sp98: updated.

PR #201, opened 2021-01-20
Bug 1917315: delete localvolumeset-local-provisoner daemonset and fix PV name hashing

rohantmp:
- Fix: PV hashing had a slight difference causing duplicate PVs (full path was being hashed as opposed to only file name in 4.6)
- Perform a one-time cleanup of the old daemonset if it exists.

### DIFF
--- a/pkg/controller/nodedaemon/reconcile.go
+++ b/pkg/controller/nodedaemon/reconcile.go
@@ -1,8 +1,17 @@
 package nodedaemon
 
 import (
+	"context"
+	"fmt"
+	"time"
+
 	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -28,9 +37,10 @@ var _ reconcile.Reconciler = &DaemonReconciler{}
 type DaemonReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client    client.Client
-	scheme    *runtime.Scheme
-	reqLogger logr.Logger
+	client                   client.Client
+	scheme                   *runtime.Scheme
+	reqLogger                logr.Logger
+	deletedStaticProvisioner bool
 }
 
 // Reconcile reads that state of the cluster for a LocalVolumeSet object and makes changes based on the state read
@@ -40,6 +50,12 @@ type DaemonReconciler struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *DaemonReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	r.reqLogger = logf.Log.WithName(controllerName).WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+
+	// do a one-time delete of the old static-provisioner daemonset
+	err := r.cleanupOldDaemonsets(request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 
 	lvSets, tolerations, ownerRefs, nodeSelector, err := r.aggregateDeamonInfo(request)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rohantmp:
Accidentally touched a previous commit when squashing the test changes, repushed.

rohantmp:
@gnufied please review :)

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -41,6 +48,24 @@ type DaemonReconciler struct {
...
+			if err != nil && !errors.IsGone(err) {
+				r.reqLogger.Error(err, fmt.Sprintf("could not delete daemonset %q", ProvisionerName))
+				return reconcile.Result{}, err
+			}
+		} else if !errors.IsNotFound(err) { // unknown error
+			r.reqLogger.Error(err, fmt.Sprintf("could not fetch daemonset %q to clean it up", ProvisionerName))
+			return reconcile.Result{}, err
+		}
+		r.deletedStaticProvisioner = true
+	}
```
gnufied, line 67: Should we wait for daemonset replicas to be 0 before reconciling the `localvolumeset`  further?

rohantmp: Is there going to be a problem if they both run https://github.com/openshift/local-storage-operator/blob/master/deploy/scripts/quick_reset.sh at the same time? Even that seems unlikely unless the controller-manager is down or something.

rohantmp: Can do though

gnufied: Yes that would be nice. I don't like the idea that we might have 2 ways of doing things running concurrently.

gnufied: There are at least couple of ways of ensuring that pods created by daemonset are torn down before we can start reconciling the `LocalVolumeSet` object further:

1. We could check for all pods created by daemonset to be deleted. I think it should be possible to do that by a selector etc.
2. There was another neat trick I saw on stackoverflow where one can set a non-existant node-selector on daemonset which would result in all pods to be deleted and then we can delete the daemonset.

I am sure there could be other ways of doing this.

rohantmp: I added a wait to confirm the pods with the label app=localvolumeset-local-provisioner are gone before setting deletedLocalProvisioner

rohantmp:
@gnufied PTAL

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -41,6 +51,12 @@ type DaemonReconciler struct {
 func (r *DaemonReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	r.reqLogger = logf.Log.WithName(controllerName).WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
+	// do a one-time delete of the old static-provisioner daemonset
+	err := r.cleanupOldDaemonsets(request.Namespace)
```
gnufied, line 55: Can you confirm if we log an event to `LocalVolumeSet` object if deleting this daemonset fails? The idea is - if somehow the `LocalVolumeSet` rollout fails after LSO upgrade, I wouldn't want customer to dig through logs to find out what happened and hence an event would be appropriate place to log this.

pkg/controller/nodedaemon/reconcile.go
```diff
@@ -67,5 +83,51 @@ func (r *DaemonReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 		r.reqLogger.Info("daemonset changed", "daemonset.Name", ds.GetName(), "op.Result", opResult)
 	}
 
-	return reconcile.Result{}, nil
+	return reconcile.Result{}, err
+}
+
+// do a one-time delete of the old static-provisioner daemonset
+func (r *DaemonReconciler) cleanupOldDaemonsets(namespace string) error {
+	if r.deletedStaticProvisioner {
```
gnufied, line 91: Is it possible that this function can be called again and again while we wait for daemonset to be deleted? I think that - we should kind of return from here or do something else - rather than spamming the API server.

rohantmp: The delete function should only be called once. Once the daemonset is gone from the api server, it simply queries for pods that the daemonset creates with reasonable intervals (even a backoff).

pkg/controller/nodedaemon/reconcile.go
rohantmp: Currently, an event reporter is not implemented in the main controller for localvolumeset. We have a refactoring task open for exporting a common event reporter package. How about we add this one once the common event reporter package is in?

I agree that any long-lasting failure should bubble up as an event, but I don't see that happening here.

rohantmp: Anyway, this is my idea for centralizing eventing: https://github.com/openshift/local-storage-operator/pull/204/files

gnufied: > I agree that any long-lasting failure should bubble up as an event, but I don't see that happening here.

I do not understand this. Customer had a working LSO in 4.6 and then they upgrade to 4.8 and LSO is stuck now. We should have clear reporting for this. Are you saying - there is no chance that DS clean up fails? There are plenty of reasons that DS cleanup may fail. For example - if any of the nodes where daemonset was running is unresponsive, then pod deletion on that node may get stuck. IMO - there are other failure scenarios too. The least we could do is, report proper errors when this happens so as customers know what to do.

gnufied:
I am going to approve this for now. I have filed separate BZ for logging the event https://bugzilla.redhat.co/show_bug.cgi?id=1925211

/lgtm

PR #204, opened 2021-02-04
 internal: add eventReporter to handle various KeyedEvents that implement the interface

rohantmp:
/hold


pkg/internal/events/disk_events.go
```diff
@@ -0,0 +1,43 @@
+package events
+
+import (
+	"fmt"
+	"os"
+)
+
+// NewDiskEvent returns a DiskEvent
+func NewDiskEvent(eventReason, message, disk, eventType string) DiskEvent {
```
rohantmp, line 9: Planning on implementing ObjectEvent or ReconcileEvent next

pkg/internal/events/disk_events.go
```diff
@@ -0,0 +1,43 @@
+package events
```
rohantmp, line 1: should probably move this to pkg/common

rohantmp:
need to rebase

rohantmp:
Need to fix the big unit test

PR #222, opened 2021-03-31
Corrects the WORKDIR of the must-gather image

huffmanca:
I've removed the trailing `/must-gather` in the WORKDIR for the must-gather image.

@yselkowitz ,

Would you mind reviewing this? I'm a little confused, as this doesn't seem to affect my local builds and testing. I can see the `gather` binary included and executable with and without this change whenever I build it locally.


yselkowitz:
> Would you mind reviewing this? I'm a little confused, as this doesn't seem to affect my local builds and testing. I can see the `gather` binary included and executable with and without this change whenever I build it locally.

What you are seeing in a local build would be the "original" `/usr/bin/gather` from must-gather, not the LSO-specific version.  However, downstream this inherits cli, not must-gather, and therefore not only do you not get the LSO `gather`, but the build fails completely:
```
chmod: cannot access '/usr/bin/gather': No such file or directory
running 'chmod +x /usr/bin/gather' failed with exit code 1
```
/lgtm

PR #227, opened 2021-04-09
Bug 1947779: Include maxUnavailable in DaemonSets

huffmanca:
Includes the `rollingUpdate` DSUpdateStrategy in the diskmaker-discovery and diskmaker-manager DaemonSets.


rohantmp:
@huffmanca I think maxUnavailable 100% would be fine

PR #228, opened 2021-04-13
Bug 1946497: Updates the way lock acquisition throws an error

huffmanca:
Updates the way lock acquisition throws an error

### DIFF
--- a/pkg/internal/diskutil.go
+++ b/pkg/internal/diskutil.go
@@ -324,14 +324,21 @@ func GetDeviceFSMap() (map[string]string, error) {
 func GetPVCreationLock(device string, symlinkDirs ...string) (ExclusiveFileLock, bool, []string, error) {
 	lock := ExclusiveFileLock{Path: device}
 	locked, err := lock.Lock()
-	if err != nil || !locked {
-		return lock, false, []string{}, err
+	// If the device is busy, then we should continue and check for symlinks
+	if err != nil && err != unix.EBUSY {
+		return lock, locked, []string{}, err
 	}
-	existingLinkPaths, err := GetMatchingSymlinksInDirs(device, symlinkDirs...)
-	if err != nil || len(existingLinkPaths) > 0 {
-		return lock, false, existingLinkPaths, err
+	existingLinkPaths, symErr := GetMatchingSymlinksInDirs(device, symlinkDirs...)
+	// If symErr is not nil, there was an error fetching the symlinks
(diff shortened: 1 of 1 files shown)


pkg/internal/diskutil.go
```diff
@@ -324,11 +324,12 @@ func GetDeviceFSMap() (map[string]string, error) {
...
 	lock := ExclusiveFileLock{Path: device}
 	locked, err := lock.Lock()
-	if err != nil || !locked {
+	if err != nil {
 		return lock, false, []string{}, err
 	}
 	existingLinkPaths, err := GetMatchingSymlinksInDirs(device, symlinkDirs...)
-	if err != nil || len(existingLinkPaths) > 0 {
+	// If we don't have any existing symlinks AND we can't get a lock, throw an error
+	if err != nil || (len(existingLinkPaths) == 0 && !locked) {
```
gnufied, line 332: But what is the error here then? if we couldn't get lock because device was busy and no existing symlinks were present then `err` variable will be nil.

huffmanca: I've updated lock acquisition to return the error regardless of if the device is busy. This keeps the same logic, but populates the error if the device is busy (which isn't reported as long as there's one or more links).

Let me know your thoughts on this approach.

pkg/internal/diskutil.go
```diff
@@ -324,13 +324,20 @@ func GetDeviceFSMap() (map[string]string, error) {
...
-	if err != nil || !locked {
+	// If the device is busy, then we should continue and check for symlinks
+	if err != nil && err != unix.EBUSY {
 		return lock, false, []string{}, err
 	}
-	existingLinkPaths, err := GetMatchingSymlinksInDirs(device, symlinkDirs...)
-	if err != nil || len(existingLinkPaths) > 0 {
+	existingLinkPaths, symErr := GetMatchingSymlinksInDirs(device, symlinkDirs...)
+	// If symErr is not nil, there was an error fetching the symlinks
+	if symErr != nil {
```
gnufied, line 333: Rather than returning hardcoded `true` or `false` values - should we just just return `locked`  variable? There could be a case where device is locked but `GetMatchingSymlinksInDirs` erred out for different reason and in which case caller should not assume device did not get locked.

huffmanca: This has been updated so that we're returning `locked` instead of true or false.

PR #237, opened 2021-05-17
Bug 1961120: added permissions to service monitoring

Elbehery:
This PR added permissions to operators service monitoring, allowing cluster version upgrade.

Currently the operators has `create` permission. When upgrading the cluster, it needs permission to `update` the service role.

### DIFF
--- a/deploy/rbac.yaml
+++ b/deploy/rbac.yaml
@@ -59,6 +59,9 @@ rules:
   - list
   - watch
   - create
+  - update
+  - patch
+  - delete
 - apiGroups:
   - apps
   resourceNames:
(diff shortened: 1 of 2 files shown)


deploy/rbac.yaml
```diff
@@ -59,6 +59,9 @@ rules:
   - list
   - watch
   - create
+  - update
+  - patch
+  - delete
```
gnufied, line 64: this needs to be fixed in CSV as well - https://github.com/openshift/local-storage-operator/blob/master/manifests/4.8/local-storage-operator.v4.8.0.clusterserviceversion.yaml

Elbehery: 👍🏽  🙏🏽

PR #239, opened 2021-05-25
Bug 1963899: Move device information to annotations (backport to 4.7)

rohantmp:
this is done because some device-ids are not valid label values.


rohantmp:
@jsafrane PTAL

jsafrane:
/lgtm
/approve
sorry about the delay

PR #243, opened 2021-06-07
Bug 1967444: Add openshift-user-critical priority class

Elbehery:
This PR assign **openshift-user-critical** priority class to operator's pods

### DIFF
--- a/pkg/controller/nodedaemon/daemonsets.go
+++ b/pkg/controller/nodedaemon/daemonsets.go
@@ -131,6 +131,7 @@ func getLocalProvisionerDSMutateFn(
 		ds.ObjectMeta.Annotations[dataHashAnnotationKey] = dataHash
 		ds.Spec.Template.Spec.Containers[0].Image = common.GetLocalProvisionerImage()
 		ds.Spec.Template.Spec.Containers[0].ImagePullPolicy = corev1.PullIfNotPresent
+
 		return nil
 	}
 }
@@ -195,6 +196,9 @@ func MutateAggregatedSpec(
 
 	// service account
 	ds.Spec.Template.Spec.ServiceAccountName = common.ProvisionerServiceAccount
+
+	// priority class
+	ds.Spec.Template.Spec.PriorityClassName = common.PriorityClassName
 
 	// tolerations
 	ds.Spec.Template.Spec.Tolerations = tolerations
--- a/manifests/4.8/local-storage-operator.v4.8.0.clusterserviceversion.yaml
+++ b/manifests/4.8/local-storage-operator.v4.8.0.clusterserviceversion.yaml
@@ -312,6 +312,7 @@ spec:
                   name: local-storage-operator
               spec:
                 serviceAccountName: local-storage-operator
+                priorityClassName: openshift-user-critical
                 containers:
                   - name: local-storage-operator
                     image: quay.io/openshift/origin-local-storage-operator:latest
--- a/pkg/common/daemonsets_utils.go
+++ b/pkg/common/daemonsets_utils.go
@@ -8,6 +8,10 @@ import (
 	corev1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 )
 
+const (
+	PriorityClassName = "openshift-user-critical"
+)
+
 func NodeSelectorMatchesNodeLabels(node *corev1.Node, nodeSelector *corev1.NodeSelector) (bool, error) {
 	if nodeSelector == nil {
 		return true, nil


pkg/controller/localvolumediscovery/localvolumediscovery_controller.go
```diff
@@ -36,6 +36,7 @@ var (
 
 const (
 	DiskMakerDiscovery = "diskmaker-discovery"
+	PriorityClassName  = "openshift-user-critical"
```
huffmanca, line 39: This won't work as a global constant (being imported by the `nodedaemon/daemonsets.go`, as there's a circular dependency.

I'd recommend moving this to `pkg/common/daemonsets_utils.go`, which is already imported in both locations.

Elbehery: 👍🏽  🙏🏽

pkg/controller/nodedaemon/daemonsets.go
```diff
@@ -99,6 +100,7 @@ func getDiskMakerDSMutateFn(
 		}
 		// to read /proc/1/mountinfo
 		ds.Spec.Template.Spec.HostPID = true
+		ds.Spec.Template.Spec.PriorityClassName = class
```
jsafrane, line 103: Should be `getLocalProvisionerDSMutateFn` updated too?

Maybe you can update only `MutateAggregatedSpec()`, it seems to be called from all other places that create a DaemonSet, but please check.

Elbehery: Good catch, I missed `getLocalProvisionerDSMutateFn` 

`MutateAggregatedSpec` is called from everywhere thats true, but it does not contain `deployment` to be updated, correct me if I am wrong :)

jsafrane: Please do, It's better to have a single place where the class is set.

Elbehery: @jsafrane done 👍🏽

pkg/controller/nodedaemon/daemonsets.go
```diff
@@ -196,6 +197,9 @@ func MutateAggregatedSpec(
 	// service account
 	ds.Spec.Template.Spec.ServiceAccountName = common.ProvisionerServiceAccount
 
+	// priority class
+	ds.Spec.Template.Spec.PriorityClassName = common.PriorityClassName
+
```
gnufied, line 202: This breaks deployment of LSO in plain k8s environment. If possible - I would like us to keep plain k8s deployments as supported configuration. I personally use it for development and testing, but other folks might be interested for different reasons.

Can we add priority class as a environment variable in CSV and apply the priority class only if environment variable is set?

gnufied: cc @priyanka19-98

priyanka19-98:
On deploying on plain k8s cluster: 
```
[pjiandan@localhost local-storage-operator]$ oc get ds
NAME                DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
diskmaker-manager   0         0         0       0            0           <none>          24s
 ```

```
oc describe ds

Events:
  Type     Reason        Age                 From                  Message
  ----     ------        ----                ----                  -------
  Warning  FailedCreate  14s (x13 over 34s)  daemonset-controller  Error creating: pods "diskmaker-manager-" is forbidden: no PriorityClass with name openshift-user-critical was found
```

```[pjiandan@localhost local-storage-operator]$ oc get PriorityClass
NAME                      VALUE        GLOBAL-DEFAULT   AGE
system-cluster-critical   2000000000   false            13m
system-node-critical      2000001000   false            13m
```
If I manually create the `PriorityClass:
```
[pjiandan@localhost ~]$ oc get PriorityClass
NAME                      VALUE        GLOBAL-DEFAULT   AGE
openshift-user-critical   0            false            4m46s
system-cluster-critical   2000000000   false            21m
system-node-critical      2000001000   false            21m
```
[pjiandan@localhost ~]$ oc get po
NAME                                    READY   STATUS    RESTARTS   AGE
diskmaker-manager-n5llc                 1/1     Running   0          3m44s
local-storage-operator-795c57b5-nkjfg   1/1     Running   0          14m

Events:
  Type    Reason            Age   From                  Message
  ----    ------            ----  ----                  -------
  Normal  SuccessfulCreate  16s   daemonset-controller  Created pod: diskmaker-manager-n5llc
```

PR #249, opened 2021-06-18
add service and servicemonitor and expose /metrics endpoint for the daemonsets

sp98:
This PR:
- Creates service and service monitor for diskmaker and discovery daemonset. 
- exposes `/metric` endpoint from the daemonset pods to emit metrics. 
- Adds all the required rbacs
- secures metrics endpoint with https/tls
- Adds Local Volume Discovery Metrics
  - Total discovered disks metric for Local Volume Discovery
- Adds Local Volume Set Metrics
  - Total Provisioned PVs
  - Total disks that didn't match the update the DeviceFilter in LocalVolumeSet CR
  - Total symlinks that were created by LocalVolumeSet that didn't match the updated `deviceInclusionSpec`
- Adds Local Volume Metrics 
  - Total Provisioned PVs
   - Total symlinks that were created by LocalVolume that didn't match the updated `DevicePath`
 

gnufied:
I am sorry for massive rebase. but before I review this, can we rebase?

config/rbac/monitoring/rbac.yaml
```diff
@@ -0,0 +1,57 @@
...
+metadata:
+  name: local-storage-monitor
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: local-storage-monitor
+subjects:
+- kind: ServiceAccount
+  name: local-storage-admin
+  namespace: default
```
gnufied, line 25: is this correct namespace?

localmetrics/exporter.go
```diff
@@ -0,0 +1,164 @@
...
+				{
+					Port:     portMetrics,
+					Path:     metricsPath,
+					Interval: scrapeInterval,
+				},
+			},
+		},
+	}
+	return serviceMonitor
+}
```
gnufied, line 137: lets move these static resources into `assets` directory or something similar (or go embed)

localmetrics/exporter.go
```diff
@@ -0,0 +1,164 @@
...
+						IntVal: int32(8383),
+						StrVal: "8383",
+					},
+				},
+			},
+			Selector: e.Labels,
+		},
+	}
+	return service
+}
```
gnufied, line 82: Have you tested these metrics to be available in prometheus? I think we will have to secure metrics endpoints using https too.

gnufied: Related discussion in slack - https://coreos.slack.com/archives/C3VS0LV41/p1625770508207000

localmetrics/exporter.go
```diff
@@ -0,0 +1,164 @@
...
+
+// createOrUpdateService creates service object or an error
+func (e *Exporter) createOrUpdateService() (*corev1.Service, error) {
+	service := e.getMetricsExporterService()
+	namespacedName := types.NamespacedName{Namespace: service.GetNamespace(), Name: service.GetName()}
+
+	log.Info("Reconciling metrics exporter service", "NamespacedName", namespacedName)
+
+	oldService := &corev1.Service{}
+	err := e.Client.Get(context.TODO(), namespacedName, oldService)
```
gnufied, line 92: I think we should not be using `context.TODO` everywhere. Can we refactor this code to use proper contexts?

config/rbac/monitoring/rbac.yaml
```diff
@@ -0,0 +1,57 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: local-storage-metrics
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
+---
```
gnufied, line 57: Don't we have to update CSV also for this?

opm-bundle/manifests/lso-metrics.yaml
```diff
@@ -0,0 +1,57 @@
...
+  namespace: openshift-local-storage
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: local-storage-metrics
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
+---
```
gnufied, line 57: Lets not make any direct changes to this directory.

localmetrics/exporter.go
sp98: using proper context from the daemonset reconciler.

config/rbac/monitoring/rbac.yaml
sp98: I haven't seen other operators doing this. They simple dump monitoring related role bindings into the bundle.

opm-bundle/manifests/lso-metrics.yaml
sp98: This is for the monitoring rbac rules.  Adding them the the bundle. I haven't see operators adding the monitoring related permissions in the CSV.

localmetrics/exporter.go
sp98: I have tested this on openshift. I was able to see the metrics under `monitoring>metrics`
Screenshot from 2021-07-08 15-11-22

gnufied: did you had to change the metrics gathering over TLS?

config/rbac/monitoring/rbac.yaml
gnufied: but in case of LSO our CSV is basically static and gets picked verbatim by ART pipeline.

localmetrics/exporter.go
sp98: No. I don't think I did anything around that. Just added the role bindings related to `openshift-monitoring` in the `/bundle/manifest` directory and added `openshift.io/cluster-monitoring: "true"` label to the namespace in `catalog-create-subscribe.yaml` file.

opm-bundle/manifests/lso-metrics.yaml
gnufied: which operators you are looking at? Please note - ART team tooling *only* looks at - https://github.com/openshift/local-storage-operator/tree/master/config/manifests directory as final version of manifests it needs to apply. `opm-bundle` directory is for our own hacking/testing purposes.

gnufied: Thanks for following up the discussion with ART and operator-sdk folks. So at very minimum- we need this rbac policy in `config/manifests` ?

sp98: looks like it.

localmetrics/exporter.go
sp98: using `assets` directory and go embed.  `go-embed` requires golang 1.16 version.

assets/localmetrics/service-monitor.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: ServiceMonitor
+metadata:
+  name: lso-metrics-exporter
+  namespace: default
+  labels:
+    app: lso-metrics-exporter
+spec:
+  namespaceSelector:
+    matchNames:
+      - default
```
gnufied, line 11: why these use default namespaces?

gnufied: I understand we are replacing them later on, but we should start with right namespace to begin with.

localmetrics/exporter.go
sp98: now securing metrics endpoint with https.

config/rbac/monitoring/rbac.yaml
```diff
@@ -0,0 +1,57 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: local-storage-monitor
+rules:
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
gnufied, line 12: I see a inconsistency here -  this rbac grants '*/*' permissions in `monitoring.coreos.com` group, but no such equivalent rule exists in CSV (or rbac that was copied in manifests directory). So either this rule is too broad or rule in CSV is too narrow.

sp98: You are right. This rule is too broad.  Servicemonitors are being created by LSO operator that use `local-storage-operator` service account. We have already grant rbac permissions for `monitoring.coreos.com` group to `local-storage-operator` service account in the CSV -  [here](https://github.com/openshift/local-storage-operator/blob/a3fd1c5c74da1ad40d708a563c0cc8e00fe16135/config/manifests/4.9/local-storage-operator.v4.9.0.clusterserviceversion.yaml#L209-L228 ) and [here](https://github.com/openshift/local-storage-operator/blob/a3fd1c5c74da1ad40d708a563c0cc8e00fe16135/opm-bundle/manifests/local-storage-operator.clusterserviceversion.yaml#L189-L209).  

So removing these permissions for `local-storage-admin` as its not required.

assets/localmetrics/service-monitor.yaml
sp98: using correct namespace in the template.

common/daemonsets_utils.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 const (
-	PriorityClassName = "openshift-user-critical"
+	PriorityClassName  = "openshift-user-critical"
+	kubeRBACProxyImage = "quay.io/openshift/origin-kube-rbac-proxy:latest"
```
gnufied, line 12: I think hardcoding this image will cause problems. Among other things - this will probably break in offline environments. Also `origin-kube-proxy` is an image that is part of OCP payload  and hence I wonder if we can refer to it in LSO.

testbin/setup-envtest.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
```
gnufied, line 1: why did we add this file?

internal/diskutil.go
```diff
@@ -364,6 +364,33 @@ func GetMatchingSymlinksInDirs(path string, dirs ...string) ([]string, error) {
...
+
+PathLoop:
+	for _, path := range paths {
+		for i, device := range validDevices {
+			isMatch, err := PathEvalsToDiskLabel(path, device.KName)
+			if err != nil {
+				return orphanedSymlinkDevices, err
+			}
+			if isMatch {
+				continue PathLoop
```
gnufied, line 384: why not just use `break` here? rather than using go-to style labels?

internal/diskutil.go
```diff
@@ -364,6 +364,33 @@ func GetMatchingSymlinksInDirs(path string, dirs ...string) ([]string, error) {
...
+			} else if i == len(validDevices)-1 {
+				orphanedSymlinkDevices = append(orphanedSymlinkDevices, path)
+			}
```
gnufied, line 387: this code is bit hard to read. As I understand - the else block checks if `path` was not found in `validDevices`. Can we rewrite this code like:

```golang
for _, path := range paths {
  found := false
  for _, device := range valiDevices {
    if isMatch {
      found = true
      break
   }
   if !found {
      orphanedSymlinkDevices = append(orphanedSymlinkDevices, path)
  }
}
```

Which IMO is bit easier to read?

testbin/setup-envtest.sh
sp98: it shouldn't be there. Removed `testbin` directory.

internal/diskutil.go
sp98: Thanks for the suggestion.  Updated the method.

common/daemonsets_utils.go
sp98: now using image as env similar to diskmaker and provisioner in the operator CR.

sp98: I don't completely understand your second concern about using `origin-kube-proxy`.  Should we use the upstream image?

diskmaker/controllers/lv/reconcile.go
```diff
@@ -59,6 +60,14 @@ const (
 	ComponentName = "localvolume-symlink-controller"
 )
 
+var watchNamespace string
+var nodeName string
+
+func init() {
+	nodeName = common.GetNodeNameEnvVar()
+	watchNamespace, _ = common.GetWatchNamespace()
```
gnufied, line 68: Where are we using this?

sp98: Not using the watchNamespace in LV controller. Removed.

common/daemonsets_utils.go
```diff
@@ -18,3 +18,32 @@ func NodeSelectorMatchesNodeLabels(node *corev1.Node, nodeSelector *corev1.NodeS
...
+		},
+
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      metricsServingCert,
+				MountPath: "/etc/tls/private",
+			},
+		},
+	}
+
```
gnufied, line 48: nit: can this be moved to daemonset YAML?

common/volume_definitions.go
```diff
@@ -93,4 +95,24 @@ var (
...
+
+	// DiskmakerMetricsCertVolume is the corev1.Volume definition for serving cert secret for diskmaker service
+	DiskmakerMetricsCertVolume = corev1.Volume{
+		Name: metricsServingCert,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: DiskMakerMetricsServingCert,
+			},
+		},
+	}
```
gnufied, line 117: same as above. This seems like pretty static information and can be moved to yaml directly.

common/daemonsets_utils.go
sp98: I thought about that.  But I believe we  might not need this for non openshift environments, right?  So I think we keep it separate from the deployment yaml and add it only if openshift is used. What do you say?

(PS:  we still haven't added check to disabled this for non openshift environments)

common/volume_definitions.go
sp98: same reason as mentioned [here](https://github.com/openshift/local-storage-operator/pull/249#discussion_r673306183). Let me know your views.

common/daemonsets_utils.go
gnufied: I am leaning towards, we are kinda breaking plain k8s deployment anyways, so we may as well use this opportunity to keep our code clean. ;-)

gnufied:
Lets merge this and fix left over issues in a follow up.

/lgtm
/approve

common/daemonsets_utils.go
rohantmp: Hey, sorry for the late comments. How are we breaking k8s deployment?

sp98: If I'm not wrong, then its the use of `openshift-user-critical` priorityClass (https://bugzilla.redhat.com/show_bug.cgi?id=1967444) that breaks plan k8s. This PR uses `service.beta.openshift.io/serving-cert-secret-name` to create secret for the certs which breaks it further.

PR #254, opened 2021-07-09
Added helper function for symlink cleanup

priyanka19-98:
This function will be called to clean up the symlinks created by PVs once the LV/LVSets are deleted.

### DIFF
--- a/common/symlink_utils.go
+++ b/common/symlink_utils.go
@@ -2,10 +2,15 @@ package common
 
 import (
 	"errors"
+	"fmt"
+	"os/exec"
 	"path"
 	"path/filepath"
 
 	"github.com/openshift/local-storage-operator/internal"
+	"github.com/prometheus/common/log"
+	v1 "k8s.io/api/core/v1"
+	provCommon "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 )
 
 // GetSymLinkSourceAndTarget returns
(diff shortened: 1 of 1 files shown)


priyanka19-98:
@rohan47

common/symlink_utils.go
```diff
@@ -37,3 +42,23 @@ func GetSymLinkSourceAndTarget(dev internal.BlockDevice, symlinkDir string) (str
...
+
+func GetCleanPVSymlinkFunc(runtimeConfig *provCommon.RuntimeConfig) func(pv *v1.PersistentVolume) error {
+	return func(pv *v1.PersistentVolume) error {
+		log.Infof("Removing symlink for %s", pv.ObjectMeta.Name)
+		config, ok := runtimeConfig.DiscoveryMap[pv.Spec.StorageClassName]
+		if !ok {
+			return fmt.Errorf("Unknown storage class name %s", pv.Spec.StorageClassName)
+		}
+		mountPath, err := provCommon.GetContainerPath(pv, config)
+		if err != nil {
```
sp98, line 54: this `err` is not returned.

sp98 (changes requested):
Please add more details to the ticket description and commit message. Like why this change is needed in the first place, etc.

common/symlink_utils.go
priyanka19-98: Thanks, made the change.

PR #260, opened 2021-07-15
Fix location of MG image

gnufied:


dobsonj:
new image works, so...
/lgtm

PR #261, opened 2021-07-20
Remove duplicate assets

gnufied:
Also automate creation of diskmaker and operator images


.gitignore
```diff
@@ -6,3 +6,4 @@ bin/
 cover.out
 _output
 testbin/
+opm-bundle/manifests/
```
sp98, line 9: nit: We should update the [hacking.md ](https://github.com/openshift/local-storage-operator/blob/master/HACKING.md#manual-creation-of-bundle-and-index-image)to reflect these changes in the manual creation steps.

sp98 (changes requested):
nit

hack/sync_bundle
```diff
@@ -50,14 +50,71 @@ class SyncBundle
...
+    content_lines.each do |line|
+      if line =~ /^FROM\s+(.+)/
+        sources << $1
+      end
+    end
+
+    sources_are_private = lambda do
+      all_private = true
+      sources.each { |source|
+        if source !~ /^registry/
```
rohantmp, line 88: nit: won't always hold that a registry is private if it starts with `registry`

rohantmp (commented):
Index image build works when tested, and haven't found any issues, but I'm not very familiar with the index building process. Have an OCP cluster coming up, will test the created index when it does and update.

hack/sync_bundle
gnufied: yes, so if Dockerfile has all `FROM` statements beginning with `registry.blah` - `yum install` will likely fail. It is basically a hack but sort of works in real world usage. :D

rohantmp (changes requested):
Did not error/exit when build failed, and later did not show any errors when opm wasn't installed:
```
Check for ./Dockerfile
Check for ./Dockerfile.diskmaker
Building operator image
Sending build context to Docker daemon  161.6MB
Step 1/10 : FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
Head "https://registry.ci.openshift.org/v2/ocp/builder/manifests/rhel-8-golang-1.16-openshift-4.9": unauthorized: authentication required
The push refers to repository [docker.io/rohantmp/local-storage-operator]
2d798c0dd320: Layer already exists 
771a71bac91b: Layer already exists 
5ed1c7a8f421: Layer already exists 
acb1910f17e5: Layer already exists 
8dc12475387c: Layer already exists 
9624be4353eb: Layer already exists 
72e7d306c279: Layer already exists 
```

gnufied:
1. Fixed. It will exit on error.
2. we can fix that in a follow up.
3. Should be fixed.

sp98:
Looks good to me.

rohantmp (approved):

PR #262, opened 2021-07-20
remove local provisioner daemonset, no longer used

dobsonj:
cleanup work following up on https://github.com/openshift/local-storage-operator/pull/257#discussion_r670922916
/assign @sp98 
/assign @rohantmp 
/assign @gnufied


common/names.go
```diff
@@ -10,7 +10,6 @@ import (
 
 const (
 	defaultDiskMakerImageVersion = "quay.io/openshift/origin-local-storage-diskmaker"
-	defaultProvisionImage        = "quay.io/openshift/origin-local-storage-static-provisioner"
```
sp98, line 13: We also have references to this image in some other places as well.  Like  `config/manager/manager.yaml` , `config/manifest/4.9/image-references` and in the csv files.

sp98 (changes requested):

common/names.go
```diff
@@ -56,14 +54,6 @@ const (
 	DiscoveryMetricsServingCert = "discovery-metric-serving-cert"
 )
 
-// GetLocalProvisionerImage return the image to be used for provisioner daemonset
-func GetLocalProvisionerImage() string {
-	if provisionerImageFromEnv := os.Getenv(ProvisionerImageEnv); provisionerImageFromEnv != "" {
```
sp98, line 61: `ProvisionerImageEnv` reference should be removed in the `names.go` file.

sp98 (changes requested):

common/names.go
dobsonj: Thank @sp98, updated.

common/names.go
dobsonj: Done

sp98 (approved):

PR #263, opened 2021-07-22
make LSO work on non-openshift clusters

sp98:
Following changes don't LSO to be deployed on vanila K8s.
1. Use of `kube-proxy-sidcar` to secure custom metrics endpoint
2. Use of service monitor. 

This PR 
- adds `service.beta.openshift.io/serving-cert-secret-name` annotation to daemonset service if `SECURE_METRICS_ENDPOINT` env is true
- adds kube-rbac-proxy sidecar container to lso daemonsets if `SECURE_METRICS_ENDPOINT` is true

This should help users to deploy LSO in non openshift environments as well. But they would still need Prometheus to create servicemonitors. 

### DIFF
--- a/assets/templates/localmetrics/service-monitor.yaml
+++ b/assets/templates/localmetrics/service-monitor.yaml
@@ -16,8 +16,3 @@ spec:
   - port: metrics
     path: /metrics
     interval: 1m
-    scheme: https
-    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
-    tlsConfig:
-      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: lso-metrics-exporter.openshift-local-storage.svc
--- a/assets/templates/localmetrics/service.yaml
+++ b/assets/templates/localmetrics/service.yaml
@@ -1,8 +1,6 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    service.beta.openshift.io/serving-cert-secret-name: lso-metrics-serving-cert
   name: lso-metrics-exporter
   namespace: openshift-local-storage
   labels:
@@ -13,6 +11,6 @@ spec:
   - name: metrics
     port: 8383
     protocol: TCP
-    targetPort: metrics
+    targetPort: 8383
   selector:
     app: lso-metrics-exporter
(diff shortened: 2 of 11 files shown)


gnufied:
I am torn. On one hand - I prefer to have as less of static information as possible in YAMLs on the other hand - it does create problems with supporting both OCP and vanilla k8s. @jsafrane do you have any opinions on this?

dobsonj:
> I am torn. On one hand - I prefer to have as much of static information as possible in YAMLs on the other hand - it does create problems with supporting both OCP and vanilla k8s. @jsafrane @dobsonj do you have any opinions on this?

@gnufied Yes, I prefer using static yaml files, and IMO it's more confusing to have some static sections in yaml and others hardcoded in the controllers. This is a general problem for other fields too (like https://github.com/openshift/local-storage-operator/pull/243/files#r670556624) so it would be better to have a general approach that we can extend.

Potential approaches that come to mind:
1) Maintain separate static yaml files for k8s and OCP deployments. This creates (mostly) duplicate files though with minor differences, that adds to the maintenance and testing burden.
2) Create slightly more complex templates where we stub out OCP sections that are only applied if `isPlatformOpenshift() == true`. So for example, the lines in service-monitor.yaml that are being removed could be replaced with `${SERVICE_MONITOR_METRICS_TLS_SETTINGS}`. When service-monitor.yaml is read, then we'd have to replace that variable with the yaml chunk defined in a different config file. The value could be different for OCP and k8s deployments.

But in both cases, if we don't run CI against k8s, it will still be really easy to break LSO on k8s. So maybe we need a broader discussion on that.

jsafrane:
Another options would be using go templates, which allow for very simple conditions.

sp98:
Thanks for the suggestions @dobsonj and @jsafrane . I'll try these out.

sp98:
> Another options would be using go templates, which allow for very simple conditions.

I tried go template. It looks good and should serve the purpose.  But one problem is that we already use - `[ReadFileAndReplace](https://github.com/openshift/local-storage-operator/blob/d49b4899c3a97506c0099d7361b7a39851c6bdd4/assets/assets.go#L23)` to replace placeholder texts in LSO manifests.  Using templates on top of that will be confusing and not a good approach (IMO)

One option would be to replace all existing `ReadFileAndReplace` usage with go templates. This could be better option than using both `ReadFileAndReplace` and `go templates`. 

@gnufied

dobsonj:
Closing this PR in favor of pulling these static resources out into yaml assets in https://github.com/openshift/local-storage-operator/pull/298
/close

PR #266, opened 2021-08-02
Enable PriorityClass for daemonset only when the platform is Openshift

priyanka19-98:
Dependent on https://github.com/openshift/local-storage-operator/pull/263


controllers/nodedaemon/daemonset_test.go
```diff
@@ -45,10 +46,12 @@ func TestMutateAggregatedSpecTemplates(t *testing.T) {
 		[]string{
 			"${OBJECT_NAMESPACE}", "new-namespace",
 			"${CONTAINER_IMAGE}", common.GetDiskMakerImage(),
+			"${PRIORITY_CLASS_NAME}", "",
 		},
 	)
 	assert.Nil(t, err, "ReadFile should not return an error when reading the template")
 	dsTemplate := resourceread.ReadDaemonSetV1OrDie(dsBytes)
+	fmt.Println(dsTemplate)
```
gnufied, line 54: is this intended?

priyanka19-98: removed

PR #271, opened 2021-08-20
Bug 2000837: Update RBAC api to v1 from v1beta1 (backport to 4.8)

priyanka19-98:

### DIFF
--- a/deploy/rbac.yaml
+++ b/deploy/rbac.yaml
@@ -1,6 +1,6 @@
 ---
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: local-storage-operator
 rules:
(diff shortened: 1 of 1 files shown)


gnufied:
@priyanka19-98 you can't use https://bugzilla.redhat.com/show_bug.cgi?id=1995653 for 4.8. You will have to clone the bug to 4.8 version and then use that bug for this PR.

deploy/rbac.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: local-storage-operator
 rules:
```
gnufied, line 6: While we are doing this - should we also update the CRDs?

priyanka19-98: CRDs are already updated to v1.

hardys:
[patch-manager] Approving for cherry-pick with a score of 0.20 - bz prio medium, I can see the corresponding change was made to master via #270

PR #273, opened 2021-08-23
Bug 1997122: localvolume: add event/log message if device already claimed

rohantmp:
LocalVolume device paths are explicitly specified, so if they are
already claimed, we should not fail siliently.


gnufied:
Is there a bug or something for the PR?

rohantmp:
Nope, but master's open right?

gnufied:
No. master is not open right now. Every PR needs BZ because we are in 4.9 code freeze.

rohantmp:
yup, realized minutes after commention :sweat_smile:

PR #274, opened 2021-08-29
Bug 1985082: remove hard coded namespace from prometheus rbac rules

sp98:
using `openshift-local-storage` namespace in prometheus rbac manifests fails to emit lso metrics in other namespaces.
This PR removes the namespace field from the prometheus-role-binding.yaml and prometheus-role.yaml. This enables
creating these resources in the same namespace as csv.

### DIFF
--- a/config/manifests/4.9/prometheus-role-binding.yaml
+++ b/config/manifests/4.9/prometheus-role-binding.yaml
@@ -2,7 +2,6 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: local-storage-metrics
-  namespace: openshift-local-storage
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
--- a/config/manifests/4.9/prometheus-role.yaml
+++ b/config/manifests/4.9/prometheus-role.yaml
@@ -2,7 +2,6 @@ kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: local-storage-metrics
-  namespace: openshift-local-storage
 rules:
  - apiGroups:
    - ""


main.go
```diff
@@ -158,3 +166,30 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func applyPrometheusRBAC(client client.Client, namespace string) error {
```
gnufied, line 170: Ideally we resync this stuff periodically. Do we miss any edge cases if we sync RBACs only in the beginning?

sp98: I don't think we miss any edge cases while syncing these RBACs only in the beginning. If I'm not wrong, having them in the manifests, would also means these RBACS will be applied in the beginning only (by ART) , right ?

rohantmp: If some other entity changes the RBAC, the operator would have to be restarted to fix the RBAC. I guess the potential for divergence from desired state is the lowest possible(user intervention required), and the desired state only changes with the image which would trigger a restart anyway.

One downside I can think of is if there is some transient issue with the apiserver, the pod could go into CrashLoopBackoff. ([max 5 minutes](https://github.com/kubernetes/kubernetes/blob/5f920426103085a28069a1ba3ec9b5301c19d075/pkg/kubelet/kubelet.go#L121))

rohantmp: Putting it in a reconcile loop would mitigate both of those issues, but would add another informer.

gnufied: yeah I do not like relying on pod crash for resync of RBAC rules. There should be a better way. :(

sp98: @gnufied  @rohantmp  So possible options here :
- Have a new controller to sycn prometheus rbac.  Seems like an overkill to have it just for Prometheus rbacs.
- Sync prometheus rbac in `DaemonReconciler` and `LocalVolumeDiscoveryReconciler` reconciler.  This leads to repetition. But we can't avoid that since both the discovery and lvs/lv daemonset had different reconcilers.
Thoughts?

assets/templates/rbac/prometheus-role-binding.yaml
```diff
@@ -2,7 +2,7 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: local-storage-metrics
-  namespace: openshift-local-storage
+  namespace: ${OBJECT_NAMESPACE}
```
rohantmp, line 5: what if we just put this in the manifests directory and omitted this namespace line? let this be created wherever olm wills it (I assume it'll create it in the LSO namespace anyay)

rohantmp (changes requested):
I think we should try everything to avoid dynamic RBAC :)

assets/templates/rbac/prometheus-role-binding.yaml
gnufied: if that works, sure. How does it behave with upgrades? say I upgraded LSO from 4.8 (where namespace is `local-storage`) to 4.9 (in same namespace).

sp98:
checking on why the e2e is failing.

assets/templates/rbac/prometheus-role-binding.yaml
sp98: worth trying.

sp98: This actually works.  I removed the namespace in the prometheus-role-binding and prometheus-role .  Tried deploying via the catalog source using [this](https://github.com/openshift/local-storage-operator/blob/master/HACKING.md#for-developing-on-openshift-and-olm).

Both of these resources were deployed in the same namespace as csv.

rohantmp: > How does it behave with upgrades?

I don't think olm behaviour on which namespace it places it in should change across upgrades

PR #275, opened 2021-08-31
Bug 1994035: increase the polling interval to 1m for lv diskmaker

rohantmp:

### DIFF
--- a/diskmaker/controllers/lv/reconcile.go
+++ b/diskmaker/controllers/lv/reconcile.go
@@ -49,7 +49,7 @@ import (
 // It also ensures that only stable device names are used.
 
 var (
-	checkDuration = 5 * time.Second
+	checkDuration = 60 * time.Second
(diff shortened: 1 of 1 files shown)


diskmaker/controllers/lv/reconcile.go
```diff
@@ -49,7 +49,7 @@ import (
 // It also ensures that only stable device names are used.
 
 var (
-	checkDuration = 5 * time.Second
+	checkDuration = 60 * time.Second
 	diskByIDPath  = "/dev/disk/by-id/*"
```
gnufied, line 53: Did we had chance to measure impact of the change on CPU usage of diskmaker pods btw?

gnufied: Roughly speaking I would expect some dip in attached chart which is available in OCP dashboard. 
Screenshot from 2021-09-01 21-47-31

PR #276, opened 2021-08-31
Bug 1993931: Storage operators use older kubernetes client

dobsonj:
https://bugzilla.redhat.com/show_bug.cgi?id=1993931
/cc @openshift/storage


dobsonj:
/hold
The unit test caught an inconsistency... https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner needs to be bumped, but the latest version of that package still relies on k8s 1.21

dobsonj:
/unhold
Just had to override the version for k8s.io/kubernetes and k8s.io/pod-security-admission.
static-provisioner should be bumped too after we have a new release there.

PR #278, opened 2021-09-10
Bug 2001765: deleter: Don't call reconciler with PVs

jsafrane:
`For()` will call `DeleteReconciler` with PVs, whose namespace is empty. This results in errors:

```
  ERROR	deleter	could not get provisioner configmap	{"Request.Namespace": "", "Request.Name": "local-pv-26ba56f", "error": "ConfigMap \"local-provisioner\" not found"}
```

PVs are already processed by `Watches()` and requests will be reconciled with the correct LSO namespace. Use the deleter ConfigMap in `For()`.


jsafrane:
/hold
A reconciler *must* have some `For()`.

sp98:
LGTM

PR #279, opened 2021-09-13
Bug 2001605: Don't block informer threads

jsafrane:
Don't sleep in the informer threads. When multiple PVs are released at the same time, one PV is processed in every 10 seconds. This further adds up with PV updates and it may take several minutes (hours!) to process all released PVs.

### DIFF
--- a/diskmaker/controllers/deleter/reconcile.go
+++ b/diskmaker/controllers/deleter/reconcile.go
@@ -220,8 +220,11 @@ func handlePVChange(runtimeConfig *provCommon.RuntimeConfig, pv *corev1.Persiste
 	}
 	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ownerNamespace}})
 	if isDelete {
-		time.Sleep(time.Second * 10)
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ownerNamespace}})
+		// Don't block the informer goroutine.
+		go func () {
+			time.Sleep(time.Second * 10)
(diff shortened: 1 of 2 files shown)


sp98:
LGTM

diskmaker/controllers/deleter/reconcile.go
```diff
@@ -220,8 +220,11 @@ func handlePVChange(runtimeConfig *provCommon.RuntimeConfig, pv *corev1.Persiste
 	}
 	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ownerNamespace}})
 	if isDelete {
-		time.Sleep(time.Second * 10)
-		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ownerNamespace}})
```
gnufied, line 224: Change itself looks good but I am doubtful about purpose of random sleep here. If I understand correctly the sleep was introduced so as cleanup status tracker inside static-provisioner has chance to mark status of PV when it starts cleaning it up. So what happens if we don't put sleep here? Is it racy? Will it cause second cleaning up of PV to be scheduled? reconciler is currently single threaded, so I am not sure how that affects.

gnufied: And why 10s? is it enough for status cleaner to catchup?

gnufied: cc @rohantmp @priyanka19-98

gnufied:
lets merge this for now and make proper design in follow up.

/lgtm

rohantmp:
Just back from PTO, will send a follow-up

PR #285, opened 2021-09-28
Bug 2007684: Don't block informer threads [4.9]

gnufied:
Don't sleep in the informer threads. When multiple PVs are released at the
same time, one PV is processed in every 10 seconds. This further adds up
with PV updates and it may take several minutes to process all released
PVs.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2007684


bertinatto (commented):
/lgtm

PR #291, opened 2021-11-02
Bug 2019452: Do not recursively add key/value pairs to logger objects

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2019452


gnufied:
Also cc @sp98 @rohantmp

dobsonj:
@gnufied There are a couple other places that need to be fixed too:
https://github.com/openshift/local-storage-operator/blob/496f4d6652b18a6a38d414a729e53a2580f1d416/diskmaker/controllers/lv/reconcile.go#L210-L211
https://github.com/openshift/local-storage-operator/blob/496f4d6652b18a6a38d414a729e53a2580f1d416/diskmaker/controllers/lvset/reconcile.go#L52-L53

gnufied:
Hmm, I think source of log lines is also not quite right.

gnufied:
I filed a separate BZ for fixing source location - https://bugzilla.redhat.com/show_bug.cgi?id=2019532

PR #292, opened 2021-11-02
Bug 2010174: Fix duplicate PVs for disk by-names

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2010174

### DIFF
--- a/diskmaker/controllers/lv/reconcile.go
+++ b/diskmaker/controllers/lv/reconcile.go
@@ -73,6 +73,22 @@ type DiskLocation struct {
 	blockDevice  internal.BlockDevice
 }
 
+type LocalVolumeReconciler struct {
+	Client          client.Client
+	Scheme          *runtime.Scheme
+	Log             logr.Logger
+	symlinkLocation string
+	localVolume     *localv1.LocalVolume
+	eventSync       *eventReporter
+	cacheSynced     bool
+
+	// static-provisioner stuff
+	cleanupTracker *provDeleter.CleanupStatusTracker
+	runtimeConfig  *provCommon.RuntimeConfig
+	deleter        *provDeleter.Deleter
+	firstRunOver   bool
+}
+
 func (r *LocalVolumeReconciler) createSymlink(
 	deviceNameLocation DiskLocation,
 	symLinkSource string,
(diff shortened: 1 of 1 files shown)


diskmaker/controllers/lv/reconcile.go
```diff
@@ -143,13 +168,38 @@ func (r *LocalVolumeReconciler) createSymlink(
...
+
+	hostname, found := nodeLabels[corev1.LabelHostname]
+	if !found {
+		return false
+	}
+
+	deviceName := filepath.Base(deviceNameLocation.diskNamePath)
+
+	for i := range pvs {
+		pv := pvs[i]
```
dobsonj, line 191: Minor nit: you could use `for _, pv := range pvs` like you did on line 276 and avoid the extra `pv := pvs[i]` line. Purely optional.

diskmaker/controllers/lv/reconcile.go
```diff
@@ -210,6 +260,30 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to initialize PV cache: %w", err)
+		}
+		for _, pv := range pvList.Items {
+			// skip non-owned PVs
+			name, found := pv.Annotations[provCommon.AnnProvisionedBy]
+			if !found || name != r.runtimeConfig.Name {
+				continue
+			}
+			addOrUpdatePV(r.runtimeConfig, pv)
```
dobsonj, line 282: When we're populating the cache here, is it possible to get a competing `addOrUpdatePV` call via `CreateFunc`?
https://github.com/openshift/local-storage-operator/blob/3defd5eb65f5e36e0009f5775225965eeecfeb11/diskmaker/controllers/lv/reconcile.go#L678-L683
If so, what would happen in that case? Could we end up with the same PV added twice?

diskmaker/controllers/lv/reconcile.go
gnufied: I was doing that on purpose to avoid globbering of loop variables by goroutines. I know that is not the case here, but I am trying to adopt this pattern as matter of style rather than only when needed.

dobsonj: Ok that's fine, just caught my eye.

diskmaker/controllers/lv/reconcile.go
gnufied: The PVs in cache are keyed by volume names, which is supposed to be a unique string cluster-wide and hence last update will win I think. For our purpose here, it won't matter I think.

PR #299, opened 2021-12-03
Bug 2019532: Logger object in LSO does not log source location accurately

dobsonj:
See https://bugzilla.redhat.com/show_bug.cgi?id=2019532
/cc @openshift/storage


dobsonj:
sample log file from these changes after adding a couple of LV's and enabling discovery:
[local-storage-operator-74b94cc95f-ssbjs.20211202.log](https://github.com/openshift/local-storage-operator/files/7645973/local-storage-operator-74b94cc95f-ssbjs.20211202.log)

dobsonj:
I ended up stashing the NewDelegatingLogger implementation on a separate branch, if that seems better:
https://github.com/openshift/local-storage-operator/compare/master...dobsonj:bug2019532
Still, it's just a workaround, and klog may be more reliable / less likely to break, and it would be more consistent with our other controllers.

dobsonj:
/hold
for manual testing

dobsonj:
/unhold
logs after the rebase:
[20220203_2019532-diskmaker-discovery-lx94c.log](https://github.com/openshift/local-storage-operator/files/7998856/20220203_2019532-diskmaker-discovery-lx94c.log)
[20220203_2019532-diskmaker-manager-k76dl.log](https://github.com/openshift/local-storage-operator/files/7998857/20220203_2019532-diskmaker-manager-k76dl.log)
[20220203_2019532-local-storage-operator-6cf5b648d4-nvwrh.log](https://github.com/openshift/local-storage-operator/files/7998858/20220203_2019532-local-storage-operator-6cf5b648d4-nvwrh.log)

common/provisioner_utils.go
```diff
@@ -233,7 +230,7 @@ func CreateLocalPV(
 		return nil
 	})
 	if opRes != controllerutil.OperationResultNone {
-		pvLogger.Info("pv changed", "operation", opRes)
+		klog.Infof("PV changed, name = %s, status = %s", pvName, opRes)
```
gnufied, line 233: Could we use `InfoS` in these places?

dobsonj: klog doesn't have any structured logging functions that I could find... at least not this version of it. Am I missing something obvious?

dobsonj: Ah, that was added in klog v2 it looks like, and LSO is still on v1:
```
$ grep klog go.mod 
        k8s.io/klog v1.0.0
```

gnufied: hmm okay. Lets handle this later.

dobsonj: Hold on, I made the changes, but need to retest.

dobsonj:
Using structured logging:
[20220203_2019532-diskmaker-discovery-gzcnf.log](https://github.com/openshift/local-storage-operator/files/7999409/20220203_2019532-diskmaker-discovery-gzcnf.log)
[20220203_2019532-diskmaker-manager-6cghw.log](https://github.com/openshift/local-storage-operator/files/7999410/20220203_2019532-diskmaker-manager-6cghw.log)
[20220203_2019532-local-storage-operator-6cf5b648d4-md5nz.log](https://github.com/openshift/local-storage-operator/files/7999411/20220203_2019532-local-storage-operator-6cf5b648d4-md5nz.log)

common/provisioner_utils.go
dobsonj: Pushed.

diskmaker/controllers/lv/reconcile.go
```diff
@@ -345,8 +340,8 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 
 	err = os.MkdirAll(r.symlinkLocation, 0755)
 	if err != nil {
-		logger.Error(err, " error creating local-storage directory ", "symLinkLocation", r.symlinkLocation)
-		os.Exit(-1)
+		klog.ErrorS(err, "error creating local-storage directory", "symLinkLocation", r.symlinkLocation)
+		return ctrl.Result{}, err
```
dobsonj, line 344: This is buried in the logging changes, so I just want to point out the `os.Exit` call I removed here--that seemed very wrong to me.

gnufied: I distinctly remember spotting this during review, but at that time my thinking was - if we can't create local-storage root directory, something is indeed wrong with the configuration/node and exiting here will cause pod to crash loop and throw easy to spot errors that might be helpful (basically complain as loudly as we can, if things aren't really working).

PR #300, opened 2021-12-08
Bug 2028685: LSO repeatedly reports errors while diskmaker-discovery pod is starting

RomanBednar:
PR contains two commits, one for documenting how to run LSO locally which can be useful for verification of some bug fixes (like this one) and second with an actual fix.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -6,6 +6,37 @@
 
 2. Create images as documented below
 
+## Running the operator locally
+
+1. Install LSO via OLM/OperatorHub in GUI
+
+2. Build the operator
+```
+make build-operator
+```
+
+3. Export env variables required by LSO [deployment](https://github.com/openshift/local-storage-operator/blob/8fc42cc8b990907c88a6da551dc85b55c2dc4417/config/manifests/4.10/local-storage-operator.clusterserviceversion.yaml#L363)
+```
+export DISKMAKER_IMAGE=quay.io/openshift/origin-local-storage-diskmaker:latest
+export KUBE_RBAC_PROXY_IMAGE=quay.io/openshift/origin-kube-rbac-proxy:latest
+export PRIORITY_CLASS_NAME=openshift-user-critical
+```
+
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -6,6 +6,42 @@
...
+
+6. Run the operator locally
+```
+~> ./_output/bin/local-storage-operator -kubeconfig=$KUBECONFIG
+```
+
+**NOTE** 
+The diskmaker-discovery pods fail to start until [BZ2027008](https://bugzilla.redhat.com/show_bug.cgi?id=2027008) is fixed
+
+
```
dobsonj, line 44: Can you remove this note? Notes about specific bugs will (hopefully) quickly become obsolete.

HACKING.md
```diff
@@ -6,6 +6,42 @@
...
+## Running the operator locally
+
+1. Install LSO via OLM/OperatorHub in GUI
+
+2. Build the operator
+```
+make build-operator
+```
+
+3. Export env variables required by LSO [deployment](https://github.com/openshift/local-storage-operator/blob/master/config/manifests/4.10/local-storage-operator.clusterserviceversion.yaml#L363)
```
dobsonj, line 18: Can you make this a permalink, so the link doesn't get broken when we move these to 4.11?

dobsonj (commented):
I'm okay with the fix, but couple of suggestions on the notes.

HACKING.md
dobsonj: Filed a PR at https://github.com/openshift/local-storage-operator/pull/301

RomanBednar: Done, thanks for the PR.

HACKING.md
RomanBednar: Good point - changed.

dobsonj:
Looks good, thanks for the changes. Now we just need https://github.com/openshift/local-storage-operator/pull/301 to merge so e2e-operator will pass.

PR #308, opened 2022-01-18
Bug 2006698: increase the polling interval to 1m for lv diskmaker

RomanBednar:
This is a backport of #275 to 4.8


RomanBednar:
Tested with 4.8 - I was able to observe a cpu usage drop on diskmaker pod:

cpu_usage

Before the patch was applied the usage was slightly over 100 milicores:

cpu_usage_pre

PR #313, opened 2022-01-20
Bug 2041620: bundle CSV alm-examples does not parse

RomanBednar:
Need to remove the trailing comma to pass operator-sdk validator check.

This check is planned to be run in CI.


dobsonj:
We can't actually merge changes into the release-4.11 branch yet, due to:
https://github.com/openshift/local-storage-operator/issues/12
And it's not needed, changes in master get pulled into release-4.11 automatically until the branch gets unfrozen:
https://github.com/openshift/local-storage-operator/blob/dd283e34202a69af4c1c75bb936eabbd501747d7/config/manifests/4.10/local-storage-operator.clusterserviceversion.yaml#L19

PR #318, opened 2022-02-08
Bug 2012111: Stop enqueing LocalVolumeSet after NotFound error

jsafrane:
Return `reconcileError` from `addAvailabilityConditions` to stop enqueing LocalVolumeSet after NotFound error.

`reconcileError` should be already `nil` if related LocalVolumeSet does not exist, unless something serious happened when cleaning up the object (and in that case requeue is expected).


jsafrane:
cc @openshift/storage

PR #328, opened 2022-03-03
Bug 2052618:Use device location specified by user for LocalVolume PV creation

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2052618

Note for reviewers:

I tested this code by disabling udev rules that create symlink in `/dev/disk/by-id` path and it seems to have worked.

### DIFF
--- a/diskmaker/controllers/lv/reconcile.go
+++ b/diskmaker/controllers/lv/reconcile.go
@@ -86,6 +86,7 @@ type LocalVolumeReconciler struct {
 	cleanupTracker *provDeleter.CleanupStatusTracker
 	runtimeConfig  *provCommon.RuntimeConfig
 	deleter        *provDeleter.Deleter
+	fsInterface    FileSystemInterface
 	firstRunOver   bool
 }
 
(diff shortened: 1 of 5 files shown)


diskmaker/controllers/lv/reconcile.go
```diff
@@ -502,6 +512,16 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+func getSymlinkSourceAndTarget(devLocation DiskLocation, symlinkDir string) (string, string, bool, error) {
+	if devLocation.diskID != "" {
+		target := path.Join(symlinkDir, filepath.Base(devLocation.diskID))
+		return devLocation.diskID, target, true, nil
+	} else {
+		target := path.Join(symlinkDir, filepath.Base(devLocation.userProvidedPath))
+		return devLocation.userProvidedPath, target, false, nil
+	}
+}
+
```
dobsonj, line 524: Should this method replace `common.GetSymLinkSourceAndTarget()` in `common/symlink_utils.go`?
I see `common.GetSymLinkSourceAndTarget()` only used one other place, for lvset. Do we need the same change there?
https://github.com/openshift/local-storage-operator/blob/a9d477a312fa3a7411366f8d025dec14b0aee8ae/diskmaker/controllers/lvset/reconcile.go#L159

gnufied: So criteria of device selection in LVSet is much stricter and different and not to mention there is no `userSpecifiedDevicePath` there and hence I left that code alone. That code will default to deviceName (such as `/dev/sda`) even if corresponding `/dev/disk/by-path` is available but `/dev/disk/by-id` is not available. It is not ideal for LVSets but just replacing `GetSymlinkSourceTarget` will not help and will require much bigger change I think.

PR #332, opened 2022-03-08
Bug 2052071: Delete obsolete ServiceMonitor

jsafrane:
ServiceMonitor `local-storage-operator-metrics` was created by LSO in 4.8 to capture metrics from LSO itself. We stopped creating the ServiceMonitor in 4.9 and LSO does not even expose the metrics port, therefore delete it.

@openshift/storage


jsafrane:
Fixed compilation error

PR #333, opened 2022-03-08
Bug 2061995: Use device location provided by the user

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2061995

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.11 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 AS builder
 WORKDIR /go/src/github.com/openshift/local-storage-operator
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.11 AS builder
```
jsafrane, line 1: It should use 4.10 images.
Perhaps we should delete unused Dockerfiles (without .rhel7 suffix) eventually.

gnufied: oops sorry I just blindly cherry-picked. yeah we could delete these. currently `sync_bundle` script refers those images and that needs to be updated too.

gnufied: fixed

gnufied:
@chao007 can you add cherry-pick approved?

/assign @chao007

PR #340, opened 2022-04-01
[release-4.10] Bug 2067702: Remove unused dependencies from go.mod

bertinatto:
Both moby and goautoneg are listed in go.mod via a replace directive, but they are no longer necessary.

Moby was added there because it used to be a dependency for helm, which was a dependency for operator-sdk.
Goautoneg was added there to override the one specified in library-go's go.mod file (which was invalid).

CC @openshift/storage


dobsonj:
/assign @duanwei33 
for cherry-pick-approved

PR #346, opened 2022-04-12
Bug 2060532: LSO e2e tests are run against default image and namespace

dobsonj:
This PR:
- Removes the code that generates custom install manifests during the e2e tests. We really don't need to maintain separate manifests and associated code to do that. The install is already handled by OLM (see https://github.com/openshift/release/pull/26479 for how the operator gets installed now).
- Removes some misc dead code. For example, there were some comments to remove `namespace` from context.go, and that field was set to the same value as `operatorNamespace` anyway. So I just removed that chunk of code and updated the callers to use `GetOperatorNamespace` instead of `GetNamespace`.
- Sets the default namespace to `openshift-local-storage`, where OLM installs the operator.
- Sets the `ServiceAccountName` field in `cleanup_hostdirs.go` to resolve a [test failure](https://github.com/openshift/local-storage-operator/pull/346#issuecomment-1097399744).

https://bugzilla.redhat.com/show_bug.cgi?id=2060532
/cc @openshift/storage

### DIFF
--- a/hack/test-e2e.sh
+++ b/hack/test-e2e.sh
@@ -1,53 +1,14 @@
 #!/bin/bash -x
 set -e
 
-ARTIFACT_DIR=${ARTIFACT_DIR:-_output}
-manifest=${ARTIFACT_DIR}/manifest.yaml
-global_manifest=${ARTIFACT_DIR}/global_manifest.yaml
-rm -f $manifest $global_manifest
-mkdir -p ${ARTIFACT_DIR}
-
-if [ -n "${IMAGE_FORMAT}" ] ; then
-    echo "IMAGE_FORMAT set as '${IMAGE_FORMAT}'"
-    IMAGE_LOCAL_STORAGE_OPERATOR=$(sed -e "s,\${component},local-storage-operator," <(echo $IMAGE_FORMAT))
-    IMAGE_LOCAL_DISKMAKER=$(sed -e "s,\${component},local-storage-diskmaker," <(echo $IMAGE_FORMAT))
-else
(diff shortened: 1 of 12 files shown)


dobsonj:
/retest
timed out waiting for cleanup jobs to complete... not clear if this is consistent

dobsonj:
> /retest timed out waiting for cleanup jobs to complete... not clear if this is consistent

The pods from the cleanup jobs failed to start:
```
  Warning  FailedCreate  35s (x4 over 105s)  job-controller  Error creating: pods "cleanup-ip-10-0-213-206.us-east-2.compute.internal-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[1]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```
Looks like we just need to add `ServiceAccountName` when creating the job in `cleanup_hostdirs.go`.

PR #350, opened 2022-04-21
Transition LSO manifests to stable channel

dobsonj:
This PR transitions LSO to using the stable channel as enabled by [ART-3107](https://issues.redhat.com/browse/ART-3107). This was already done in [ptp-operator](https://github.com/openshift/ptp-operator/tree/release-4.10/manifests), so I followed that example.
This saves us from some backporting headaches and from renaming files every release, but it seems we'll still have to update the versions in the CSV for example each release.
/cc @openshift/storage


jsafrane:
/lgtm
/approve
@duanwei33 @chao007 please check how things will look like on your end. Goal: have just channel named 'stable' in OLM and not to have '4.11'.

chao007:
@jsafrane Thanks very much for this info.

duanwei33:
cc @Phaow To see if it will impact our automation.

PR #357, opened 2022-06-29
Bug 2091873: PVs get stuck in Released after PVC is deleted and Node UID changed

RomanBednar:
cc @openshift/storage 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2091873

The patch changes PV matching in deleter reconcile loop so it will match also PVs that have node UID in the annotation, effectively ignoring the UID in the cleanup/delete flow. This is needed because node UID is not persistent and if a node gets deleted the UID changes. This can not work because reconcile tired to do exact matching and so PVs were getting stuck in Released state when UID changed.

Manually verified with following steps:

1) PV is bound (with annotation containing random UID)
```
$ oc get pv
NAME                CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM           STORAGECLASS   REASON   AGE
local-pv-56c29ba0   10Gi       RWO            Delete           Bound    default/mypvc   mylvs                   80s

$ oc get pv/local-pv-56c29ba0 -o json | jq '.metadata.annotations["pv.kubernetes.io/provisioned-by"]'
"local-volume-provisioner-ip-10-0-135-68.ec2.internal-dca4aa94-fb64-4803-b576-0a8b7019a86c"
```

2) Delete deployment and PVC to release the PV:
```
$ oc delete pvc/mypvc deployment/myapp
persistentvolumeclaim "mypvc" deleted
deployment.apps "myapp" deleted
```

3) Verify PV is Available again and not stuck in Released state:
```
$ oc get pv
NAME                CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM   STORAGECLASS   REASON   AGE
local-pv-56c29ba0   10Gi       RWO            Delete           Available           mylvs                   7s
```

4) Check the diskmaker pod to see that we log info about UID being ignored

`I0708 13:33:35.486306  303623 names.go:117] "PV matches provisioner name (UID is ignored)." pvName="local-pv-56c29ba0" pvAnnotation="local-volume-provisioner-ip-10-0-135-68.ec2.internal-dca4aa94-fb64-4803-b576-0a8b7019a86c" provisionerName="local-volume-provisioner-ip-10-0-135-68.ec2.internal" UID="dca4aa94-fb64-4803-b576-0a8b7019a86c"`

5) Check that LSO re-created PV without UID in annotation
```
$ oc get pv/local-pv-56c29ba0 -o json | jq '.metadata.annotations["pv.kubernetes.io/provisioned-by"]'
"local-volume-provisioner-ip-10-0-135-68.ec2.internal"
```

### DIFF
--- a/common/names.go
+++ b/common/names.go
@@ -2,7 +2,10 @@ package common
 
 import (
 	"fmt"
+	"k8s.io/klog/v2"
 	"os"
+	"regexp"
+	provCommon "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 
 	localv1 "github.com/openshift/local-storage-operator/api/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


diskmaker/controllers/deleter/reconcile.go
```diff
@@ -87,6 +88,11 @@ func (r *DeleteReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{}, fmt.Errorf("failed to initialize PV cache: %w", err)
 		}
 		for _, pv := range pvList.Items {
+			err = clearUIDFromPV(&pv, r.Client, r.runtimeConfig)
```
jsafrane, line 91: What if `node.Name` looks like an UID? You will delete it here. IMO there should be a generic function `pvMatchesNode(pv, node)`, that would check the annotation is either `local-volume-provisioner-NODE_NAME` **or** its starts with `local-volume-provisioner-NODE_NAME` and the rest looks like an UUID.

And I'd use the fuction everywhere where the annotation is evaluated and current node matched, not only in the deleter. I would not save the PV in the API server, but it's matter of style, it's not bad to save it. Still, diskmaker should only update PVs it owns, not PVs on the other nodes.

RomanBednar: Thanks for the review, posting new version.

common/names.go
```diff
@@ -87,5 +90,34 @@ func LocalVolumeKey(lv *localv1.LocalVolume) string {
...
 // the key is provCommon.AnnProvisionedBy ("pv.kubernetes.io/provisioned-by")
 func GetProvisionedByValue(node corev1.Node) string {
-	return fmt.Sprintf("local-volume-provisioner-%v-%v", node.Name, node.UID)
+	return fmt.Sprintf("local-volume-provisioner-%v", node.Name)
+}
+
+func PVMatchesProvisioner(pv corev1.PersistentVolume, provisionerName string) bool {
+	PVAnnotation, found := pv.Annotations[provCommon.AnnProvisionedBy]
+	if !found {
+		klog.InfoS("PV annotation not found - skipping.", "pvName", pv.GetName(), "provisionerName", provisionerName)
```
dobsonj, line 99: Since this function `PVMatchesProvisioner` is called by several Reconcile loops, I think it's a bad idea to call `InfoS` in each case... this seems like it would flood the logs over time.

dobsonj (commented):
I think the fix makes sense, just one minor suggestion.
Can you also summarize what the fix does in your PR description?

common/names.go
RomanBednar: Ack - the logs seemed already quite flooded but good point, this does not have to be info level. We could increase the log level (see current version of the patch). That should be safe because I don't see -v flag being used for the deployment and default should be lower than 4.

That's what I initially wanted to do. However, I noticed that diskmaker (binary) does not seem to support changing log levels atm. That's probably why it's not in the deployment/daemonset. Maybe I'm missing something?

Or we can just remove these messages.

dobsonj: Thanks, I think setting these to log level 4 makes sense.
The reason logLevel is not propagated to diskmaker is due to https://bugzilla.redhat.com/show_bug.cgi?id=2058001 -- I haven't had time to work on it, feel free to steal that bug if you get bored :)

dobsonj:
/lgtm
@RomanBednar do you still need the `hold` tag?

/assign @jsafrane @gnufied 
for approval

RomanBednar:
Thank you, hold is no longer needed.

/unhold

RomanBednar:
> I think the fix makes sense, just one minor suggestion. Can you also summarize what the fix does in your PR description?

Added.

PR #363, opened 2022-07-18
STOR-977: Refactor reconcilers

RomanBednar:
Introduce a new helper function - ReloadRuntimeConfig() that loads runtime config values needed for reconcile loops at one place. This should help make the code cleaner by removing copy-paste code and scattered runtime config changes.

Also adding proper constructors to reconcilers which should improve readability and ease constructing them with fakes in tests.

cc @openshift/storage

### DIFF
--- a/common/controller_runtime_utils.go
+++ b/common/controller_runtime_utils.go
@@ -1,13 +1,19 @@
 package common
 
 import (
+	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
 	"os"
+	ctrl "sigs.k8s.io/controller-runtime"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	staticProvisioner "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 )
 
 // EnqueueOnlyLabeledSubcomponents returns a predicate that filters only objects that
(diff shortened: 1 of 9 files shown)


diskmaker/controllers/deleter/reconcile.go
```diff
@@ -123,29 +95,38 @@ type DeleteReconciler struct {
...
+		runtimeConfig: runtimeConfig,
+		deleter:       deleter,
+	}
+
+	return deleteReconciler
+}
+
+func (r *DeleteReconciler) WithManager(mgr ctrl.Manager) (*DeleteReconciler, error) {
+
+	r.runtimeConfig.Recorder = mgr.GetEventRecorderFor(ComponentName)
```
RomanBednar, line 127: Changing runtimeConfig on multiple places like this is not ideal, however I'm not sure what change would improve it without changing the structure too much. The problem is that the runtimeConfig contains many values that are available at different parts of the flow, most of them are obtained only after Reconcile() starts.

We could remove WithManager and make ctrl.Manager part of the constructor to make it cleaner, but I'm not sure if there might be a case where manager is not available.

More ideas are welcome.

common/controller_runtime_utils.go
```diff
@@ -61,3 +67,43 @@ func GetWatchNamespace() (string, error) {
...
+	rc.Node = node
+	rc.Namespace = request.Namespace
+	rc.DiscoveryMap = provisionerConfig.StorageClassConfig
+	rc.NodeLabelsForPV = provisionerConfig.NodeLabelsForPV
+	rc.SetPVOwnerRef = provisionerConfig.SetPVOwnerRef
+	rc.UseNodeNameOnly = provisionerConfig.UseNodeNameOnly
+	rc.MinResyncPeriod = provisionerConfig.MinResyncPeriod
+	rc.UseAlphaAPI = provisionerConfig.UseAlphaAPI
+	rc.LabelsForPV = provisionerConfig.LabelsForPV
+
```
gnufied, line 106: may as well move:

```
r.runtimeConfig.UseJobForCleaning = false
```

diskmaker/controllers/lvset/reconcile.go
```diff
@@ -98,38 +92,13 @@ func (r *LocalVolumeSetReconciler) Reconcile(ctx context.Context, request ctrl.R
...
-
-	// ignored by our implementation of static-provisioner,
-	// but not by deleter (if applicable)
-	r.runtimeConfig.UseNodeNameOnly = provisionerConfig.UseNodeNameOnly
-	r.runtimeConfig.MinResyncPeriod = provisionerConfig.MinResyncPeriod
-	r.runtimeConfig.UseAlphaAPI = provisionerConfig.UseAlphaAPI
-	r.runtimeConfig.LabelsForPV = provisionerConfig.LabelsForPV
-
 	// unsupported
 	r.runtimeConfig.UseJobForCleaning = false
```
gnufied, line 96: same here

common/controller_runtime_utils.go
RomanBednar: Done. It was not used in deleter so I did not want touch it. But since it will default to false anyway it should be safe to move it to the helper. Thanks.

diskmaker/controllers/lvset/reconcile.go
RomanBednar: Done.

jsafrane:
This is a minor refactoring, CI should catch regressions
/label docs-approved
/label px-approved

PR #371, opened 2022-10-04
OCPBUGS-2450: do not flag PV as cleaned in cache too early

RomanBednar:
Carry patch with a fix has been merged: [openshift/sig-storage-local-static-provisioner#42](https://github.com/openshift/sig-storage-local-static-provisioner/pull/42).

Bump to the fixed version and vendor.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -132,6 +132,6 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.25.1
 )
 
-replace sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb // for bug 2032924
+replace sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae //BUG: https://issues.redhat.com/browse/OCPBUGS-2450
 
 replace github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt v3.2.1+incompatible
--- a/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go
+++ b/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go
@@ -169,6 +169,8 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 		klog.Infof("Deleting pv %s after successful cleanup", pv.Name)
 		if err = d.APIUtil.DeletePV(pv.Name); err != nil {
 			if !errors.IsNotFound(err) {
+				// PV failed to delete in API server, flag the PV as not cleaned so next reconcile attempts the deletion again.
+				d.Cache.UncleanPV(pv)
 				d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete,
 					err.Error())
 				return fmt.Errorf("Error deleting PV %q: %v", pv.Name, err.Error())
--- a/go.sum
+++ b/go.sum
@@ -557,8 +557,8 @@ github.com/openshift/client-go v0.0.0-20220915152853-9dfefb19db2e h1:ab+BJg7h50p
 github.com/openshift/client-go v0.0.0-20220915152853-9dfefb19db2e/go.mod h1:e+TTiBDGWB3O3p3iAzl054x3cZDWhrZ5+jxJRCdEFkA=
 github.com/openshift/library-go v0.0.0-20220915130036-73d5a4a82865 h1:x7KWaYzkD2KQ3rha9u7OVAfjZpSFTmJRSHb4CHc+CwM=
 github.com/openshift/library-go v0.0.0-20220915130036-73d5a4a82865/go.mod h1:KPBAXGaq7pPmA+1wUVtKr5Axg3R68IomWDkzaOxIhxM=
-github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb h1:3lWe/gOwI0oMBmVkacbKA/JA6JyRnZDFxQorMtRVifs=
-github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb/go.mod h1:9IxtMY+/rQNK7aQ3MMjrcMOF8M/ej8ql9493/24p/KU=
+github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae h1:Pe4O3/6VZx9pzNfuiVuEWFYgMHdTIdpf47HlXXb4ZCg=
+github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae/go.mod h1:9IxtMY+/rQNK7aQ3MMjrcMOF8M/ej8ql9493/24p/KU=
 github.com/opentracing/opentracing-go v1.1.0/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
 github.com/pascaldekloe/goe v0.0.0-20180627143212-57f6aae5913c/go.mod h1:lzWF7FIEvWOWxwDKqyGYQf6ZUaNfKdP144TG7ZOy1lc=
 github.com/pborman/uuid v1.2.1 h1:+ZZIw58t/ozdjRaXh/3awHfmWRbzYxJoAdNJxe/3pvw=
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -979,7 +979,7 @@ sigs.k8s.io/kube-storage-version-migrator/pkg/apis/migration/v1alpha1
 # sigs.k8s.io/sig-storage-lib-external-provisioner/v6 v6.3.0
 ## explicit; go 1.13
 sigs.k8s.io/sig-storage-lib-external-provisioner/v6/util
-# sigs.k8s.io/sig-storage-local-static-provisioner v0.0.0-20220816120907-a856c67f2f68 => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb
+# sigs.k8s.io/sig-storage-local-static-provisioner v0.0.0-20220816120907-a856c67f2f68 => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae
 ## explicit; go 1.13
 sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cache
 sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common
@@ -1021,5 +1021,5 @@ sigs.k8s.io/yaml
 # k8s.io/mount-utils => k8s.io/mount-utils v0.25.1
 # k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.25.1
 # k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.25.1
-# sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb
+# sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae
 # github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt v3.2.1+incompatible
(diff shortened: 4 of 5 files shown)


jsafrane:
You must fix the issue in https://github.com/openshift/sig-storage-local-static-provisioner first and then revendor the library from there. Something like https://github.com/openshift/sig-storage-local-static-provisioner/pull/38, but without merging master from upstream.

cc @dobsonj

vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go
```diff
@@ -174,6 +173,9 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 				return fmt.Errorf("Error deleting PV %q: %v", pv.Name, err.Error())
 			}
 		}
+		// Now we're certain the PV was deleted in API and can update cache
+		d.Cache.CleanPV(pv)
```
dobsonj, line 177: I worry there may be a race in either case.
1) clean the cache entry before `d.APIUtil.DeletePV`: if the API call fails, then it won't be retried.
2) clean the cache entry after `d.APIUtil.DeletePV`: could we end up with the [original problem](https://bugzilla.redhat.com/show_bug.cgi?id=2032924) where [DeletePVs is called](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/diskmaker/controllers/deleter/reconcile.go#L71) at the wrong time and end up starting a second cleanup job before the PV is removed from the cache?

There is also a problem with this ordering, in that [DeletePV](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cache/cache.go#L78) removes the 'cleaned' cache entry, [CleanPV](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cache/cache.go#L99) adds it back as true, which means the next time LSO attempts to clean a PV with this name, it [will think it was already cleaned up successfully](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cache/cache.go#L109) and do nothing. So it causes us not to clean the PV the next time.

Is it possible to leave the call to `CleanPV` before the call to `DeletePV`, and consider the cleanup part successful, but still retry the call to `DeletePV` if the API call fails? This doesn't seem so simple, because in order to [check the CSSucceeded state](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L166) we actually [remove the cleanup job status](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L160) when checking it!
There is also the option to do a real refactoring (see [bug 2095077](https://bugzilla.redhat.com/show_bug.cgi?id=2095077)) but that might not be a good backport candidate. Still, any other option I see leads to more complexity and potential gaps.

dobsonj: > 2. clean the cache entry after `d.APIUtil.DeletePV`: could we end up with the [original problem](https://bugzilla.redhat.com/show_bug.cgi?id=2032924) where [DeletePVs is called](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/diskmaker/controllers/deleter/reconcile.go#L71) at the wrong time and end up starting a second cleanup job before the PV is removed from the cache?

I think this is not really a problem, it should not be possible to trigger `DeletePVs` more than once at a time from the `Reconcile` loop. But the ordering problem I mentioned does look like it will cause issues, at least for the current implementation.

RomanBednar:
> You must fix the issue in https://github.com/openshift/sig-storage-local-static-provisioner first and then revendor the library from there.

Sure, I should have flagged this as WIP. Doing that now.

vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go
RomanBednar: > Is it possible to leave the call to `CleanPV` before the call to `DeletePV`, and consider the cleanup part successful, but still retry the call to `DeletePV` if the API call fails? This doesn't seem so simple, because in order to [check the CSSucceeded state](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L166) we actually [remove the cleanup job status](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L160) when checking it! 

Changing something while *just checking* is strange indeed. But the cleanup job status is [added back](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L211) if it's not a success. So I think there are two options:

1) Add a function similar to [RemoveStatus()](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L388) which would just report the status and not delete it making the code more readable. But that will require some refactoring.

2) Leave the cache cleaning *before* API call but [on failure](https://github.com/openshift/local-storage-operator/blob/664054169e9d761e2d109d4c5f9bf4d7a1c4397c/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter/deleter.go#L172) add it back along with the cleanup job. 


> There is also the option to do a real refactoring (see [bug 2095077](https://bugzilla.redhat.com/show_bug.cgi?id=2095077)) but that might not be a good backport candidate. Still, any other option I see leads to more complexity and potential gaps.

That's the best but most time consuming approach. Looking at your suggestions in BZ I think option 2 (refactor to use events) would be the best. Just because option 1 means more code to maintain and option 3 reminds me of PV controllers resync approach which turned out to be "not the best".

RomanBednar:
I've tested the approach that we discussed with a slight difference that I'm adding only the cache record back on API failure, the cleanup job does not have to be recovered because at this stage we already know the cleanup was successful and don't have to run again. 

By adding the PV back to cache we enter the loop again, don't repeat the cleanup and just repeat the API deletion. @dobsonj @gnufied what do you think? Is there maybe some test we could run to rule out regressing the original race issue?

dobsonj:
> By adding the PV back to cache we enter the loop again, don't repeat the cleanup and just repeat the API deletion. @dobsonj @gnufied what do you think?

I think this looks okay.
We will end up retrying cleanup if API deletion fails (cleaned=false and no cleanup job means it will assume it was not cleaned and start a new job again from the beginning) but I don't see a big problem with that. This seems like the simplest option that does not lead to A) leaving a cleanup job status behind for a PV that has already been cleaned, or B) leaving the PV stuck in released state forever if the API call fails.

> Is there maybe some test we could run to rule out regressing the original race issue?

Can you try out the script / procedure I used to verify the original fix?
https://bugzilla.redhat.com/show_bug.cgi?id=2032924#c16
I can dig up more notes from my testing on that bug if needed.

RomanBednar:
I've ran the script for ~2 hours and was not able to observe any failure.

Resources used:
```
$ cat /tmp/pvc-pod-lv.yaml 
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: local-pv-claim1
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  volumeName: local-pv-976fb4e2
  storageClassName: lv-loop0
  volumeMode: Filesystem
---
apiVersion: v1
kind: Pod
metadata:
  name: example
  labels:
    app: example
spec:
  volumes:
  - name: claim1-vol
    persistentVolumeClaim:
      claimName: local-pv-claim1
  containers:
  - image: busybox
    command:
      - "/bin/sh"
      - "-ec"
      - "touch \"/mnt/claim1/$(date -Iseconds).txt\"; sleep 604800"
    imagePullPolicy: IfNotPresent
    name: busybox
    volumeMounts:
      - mountPath: "/mnt/claim1"
        name: claim1-vol
  restartPolicy: Always
---
apiVersion: "local.storage.openshift.io/v1"
kind: "LocalVolume"
metadata:
  name: "local-pv-976fb4e2"
  namespace: "openshift-local-storage"
spec:
  storageClassDevices:
    - storageClassName: "lv-loop0"
      volumeMode: Filesystem
      fsType: xfs
      devicePaths:
        - /dev/loop0

```

Test run:
```
$ /tmp/test.sh 
++ date +%Y
+ year=2022
+ claimdir=/mnt/claim1
+ podname=pod/example
+ pvcname=pvc/local-pv-claim1
+ oc apply -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim/local-pv-claim1 created
pod/example created
localvolume.local.storage.openshift.io/local-pv-976fb4e2 created
+ sleep 30
+ true
+ oc delete -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim "local-pv-claim1" deleted
pod "example" deleted
localvolume.local.storage.openshift.io "local-pv-976fb4e2" deleted
+ sleep 10
+ oc apply -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim/local-pv-claim1 created
pod/example created
localvolume.local.storage.openshift.io/local-pv-976fb4e2 created
+ sleep 60
+ oc get pod/example
NAME      READY   STATUS    RESTARTS   AGE
example   1/1     Running   0          61s
+ oc get pvc/local-pv-claim1
NAME              STATUS   VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS   AGE
local-pv-claim1   Bound    local-pv-976fb4e2   1Gi        RWO            lv-loop0       61s
+ oc debug pod/example -- ls -l /mnt/claim1
Starting pod/example-debug, command was: /bin/sh -ec touch "/mnt/claim1/$(date -Iseconds).txt"; sleep 604800
total 0
-rw-r--r--    1 root     root             0 Oct 13 09:51 2022-10-13T09:51:04+00:00.txt

Removing debug pod ...
++ oc debug pod/example -- ls -l /mnt/claim1
++ grep 2022
++ wc -l
+ numfiles=1
+ '[' 1 -gt 1 ']'
+ true
+ oc delete -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim "local-pv-claim1" deleted
pod "example" deleted
localvolume.local.storage.openshift.io "local-pv-976fb4e2" deleted
+ sleep 10
.
.
.
+ oc apply -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim/local-pv-claim1 created
pod/example created
localvolume.local.storage.openshift.io/local-pv-976fb4e2 created
+ sleep 60
+ oc get pod/example
NAME      READY   STATUS    RESTARTS   AGE
example   1/1     Running   0          61s
+ oc get pvc/local-pv-claim1
NAME              STATUS   VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS   AGE
local-pv-claim1   Bound    local-pv-976fb4e2   1Gi        RWO            lv-loop0       62s
+ oc debug pod/example -- ls -l /mnt/claim1
Starting pod/example-debug, command was: /bin/sh -ec touch "/mnt/claim1/$(date -Iseconds).txt"; sleep 604800
total 0
-rw-r--r--    1 root     root             0 Oct 13 11:58 2022-10-13T11:58:06+00:00.txt

Removing debug pod ...
++ oc debug pod/example -- ls -l /mnt/claim1
++ grep 2022
++ wc -l
+ numfiles=1
+ '[' 1 -gt 1 ']'
+ true
+ oc delete -f /tmp/pvc-pod-lv.yaml
persistentvolumeclaim "local-pv-claim1" deleted
pod "example" deleted
localvolume.local.storage.openshift.io "local-pv-976fb4e2" deleted
+ sleep 10
```

go.mod
```diff
@@ -132,6 +132,6 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.25.1
 )
 
-replace sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20220208224017-4561a388c9eb // for bug 2032924
+replace sigs.k8s.io/sig-storage-local-static-provisioner => github.com/openshift/sig-storage-local-static-provisioner v0.0.0-20221014091355-396d28cafaae
```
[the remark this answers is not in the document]
RomanBednar: Done. For the 4.11 and 4.10 backports: should we link to the original (this) bug as well or their clones for each branch instead?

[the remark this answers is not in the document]
jsafrane: IMO it does not really matter. As long as we can see why do we have `replace` there.

PR #376, opened 2022-11-28
Bug 2067907: bump static provisioner to v2.5.0

dobsonj:
https://bugzilla.redhat.com/show_bug.cgi?id=2067907
This pulls in the changes from https://github.com/openshift/sig-storage-local-static-provisioner/pull/46
/cc @openshift/storage


jsafrane:
The test failures look genuine.

dobsonj:
There were a couple of build failures to fix:
[NewVolumeUtil](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/a30f35b62319b9237b9c7d7531a25beac09b48d5/pkg/util/volume_util_linux.go#L38) now returns an error value that is always nil on Linux.
[GetFsCapacityByte](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/a30f35b62319b9237b9c7d7531a25beac09b48d5/pkg/util/volume_util_linux.go#L45) now requires a new `hostPath` argument that is always ignored on Linux.

PR #386, opened 2023-01-18
OCPBUGS-5004: Optimize `blkid -s TYPE` by adding list of devices to its args

mpatlasov:


dobsonj:
/retitle OCPBUGS-5004: Optimize `blkid -s TYPE` by adding list of devices to its args
I like it, looks good.
/lgtm

/assign @jsafrane
for approval and to get a second pair of eyes

dobsonj:
/jira refresh
just added target version in jira

diskmaker/controllers/lv/reconcile.go
```diff
@@ -330,8 +330,17 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		devicePaths := disks.DevicePaths
+		for _, devicePath := range devicePaths {
+			devices = append(devices, devicePath)
+		}
+	}
+
 	// list block devices
-	blockDevices, badRows, err := internal.ListBlockDevices()
+	blockDevices, badRows, err := internal.ListBlockDevices(devices)
 	if err != nil {
```
gnufied, line 344: Could this result in a regression? For example - if user specifies a non-existant device and `blkid` fails and no disk are provisoned at all? Previously - we were provisioning whatever we could find..

gnufied: Let me clarify. Say user specified following `LocalVolume`:

```yaml
...
...
devicePaths:
- /dev/vdc  # a real device
- /dev/foobar # a non-existant device
```

I am pretty sure previously we would just provision a PV out of `/dev/vdc` and the non-existent device.  *after* your change - will this result in no PV being provisioned out of `/dev/vdc` too, because whole of `blkid` will fail?

gnufied: Just tested with `blkid`- it appears that specifying non-existant device as a param simply causes it to be ignored (if there were other valid devices too) and does not result in entire command erroring out (as I was expecting). So we may be okay. 

It does mean that, if `blkid` ever changes their behaviour we will be broken.

PR #387, opened 2023-02-01
 STOR-1019: Bump to k8s 1.26 libs for OCP 4.13

bertinatto:
This patches bumps `library-go`, which in turn updates to k8s 1.26 libs:

`go get github.com/openshift/library-go && go mod tidy && go mod vendor`

This also pulls in https://github.com/openshift/library-go/pull/1460, which bumps `golang.org/x/net` v0.5.0 and addresses a CVE.

In addition to that, there are commits to also:

1. Bump Go to  1.19
2. Bump controller-runtime

CC @openshift/storage


jsafrane:
/label docs-approved
/label px-approved
/label qe-approved

This is not a user-facing change.
Also, discussed with the team offline and we agreed that CI should be enough for testing this change.

PR #394, opened 2023-04-14
OCPBUGS-10884: add multipath support for volume discovery

RomanBednar:
Adds multipath devices to discovery result.

Unrelated to discovery: we should expose multipath in console to improve user experience when creating `LocalVolumeSet` with multipath: https://github.com/openshift/console/pull/12723

cc @openshift/storage

### DIFF
--- a/diskmaker/discovery/discovery.go
+++ b/diskmaker/discovery/discovery.go
@@ -194,7 +194,26 @@ func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.Discove
 		discoveredDevices = append(discoveredDevices, discoveredDevice)
 	}
 
-	return discoveredDevices
+	return uniqueDevices(discoveredDevices)
+}
+
+// uniqueDevices removes duplicate devices from the list using DeviceID as a key
+// TODO: remove this and use lsblk with -M flag once base images are updated and lsblk v2.34 or higher is available
+// See: https://github.com/util-linux/util-linux/blob/3be31a106c52e093928afbea2cddbdbe44cfb357/Documentation/releases/v2.34-ReleaseNotes#L18
+func uniqueDevices(sample []v1alpha1.DiscoveredDevice) []v1alpha1.DiscoveredDevice {
+	var unique []v1alpha1.DiscoveredDevice
+	type key struct{ value string }
+	m := make(map[key]int)
+	for _, v := range sample {
+		k := key{v.DeviceID}
+		if i, ok := m[k]; ok {
+			unique[i] = v
+		} else {
+			m[k] = len(unique)
+			unique = append(unique, v)
+		}
+	}
+	return unique
 }
 
 // ignoreDevices checks if a device should be ignored during discovery
--- a/diskmaker/discovery/discovery.go
+++ b/diskmaker/discovery/discovery.go
@@ -179,8 +179,12 @@ func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.Discove
 			klog.Warningf("failed to parse size for the device %q. Error %v", blockDevice.Name, err)
 		}
 
+		path, err := blockDevice.GetDevPath()
+		if err != nil {
+			klog.Warningf("failed to parse path for the device %q. Error %v", blockDevice.KName, err)
+		}
 		discoveredDevice := v1alpha1.DiscoveredDevice{
-			Path:     fmt.Sprintf("/dev/%s", blockDevice.Name),
+			Path:     path,
 			Model:    blockDevice.Model,
 			Vendor:   blockDevice.Vendor,
 			FSType:   blockDevice.FSType,
@@ -308,6 +312,8 @@ func parseDeviceType(deviceType string) v1alpha1.DiscoveredDeviceType {
 		return v1alpha1.PartType
 	case deviceType == "lvm":
 		return v1alpha1.LVMType
+	case deviceType == "mpath":
+		return v1alpha1.MultiPathType
 	}
 
 	return ""
--- a/diskmaker/discovery/discovery.go
+++ b/diskmaker/discovery/discovery.go
@@ -206,10 +206,10 @@ func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.Discove
 // See: https://github.com/util-linux/util-linux/blob/3be31a106c52e093928afbea2cddbdbe44cfb357/Documentation/releases/v2.34-ReleaseNotes#L18
 func uniqueDevices(sample []v1alpha1.DiscoveredDevice) []v1alpha1.DiscoveredDevice {
 	var unique []v1alpha1.DiscoveredDevice
-	type key struct{ value string }
+	type key struct{ value, value2 string }
 	m := make(map[key]int)
 	for _, v := range sample {
-		k := key{v.DeviceID}
+		k := key{v.DeviceID, v.Path}
 		if i, ok := m[k]; ok {
 			unique[i] = v
 		} else {
--- a/diskmaker/discovery/discovery.go
+++ b/diskmaker/discovery/discovery.go
@@ -147,12 +147,9 @@ func (discovery *DeviceDiscovery) discoverDevices() error {
 
 // getValidBlockDevices fetchs all the block devices sutitable for discovery
 func getValidBlockDevices() ([]internal.BlockDevice, error) {
-	blockDevices, badRows, err := internal.ListBlockDevices([]string{})
+	blockDevices, output, err := internal.ListBlockDevices([]string{})
 	if err != nil {
-
-		return blockDevices, errors.Wrapf(err, "failed to list all the block devices in the node.")
-	} else if len(badRows) > 0 {
-		klog.Warningf("failed to parse all the lsblk rows. Bad rows: %+v", badRows)
+		return blockDevices, errors.Wrapf(err, "failed to list all the block devices in the node, stderr=%v", output)
 	}
 
 	// Get valid list of devices
--- a/api/v1alpha1/localvolumediscovery_types.go
+++ b/api/v1alpha1/localvolumediscovery_types.go
@@ -43,6 +43,8 @@ const (
 	PartType DiscoveredDeviceType = "part"
 	// LVMType is an LVM type
 	LVMType DiscoveredDeviceType = "lvm"
+	// MultiPathType is a multipath type
+	MultiPathType DiscoveredDeviceType = "mpath"
 )
 
 // LocalVolumeDiscoverySpec defines the desired state of LocalVolumeDiscovery
--- a/diskmaker/controllers/lvset/matcher_test.go
+++ b/diskmaker/controllers/lvset/matcher_test.go
@@ -395,6 +395,12 @@ func TestInTypeList(t *testing.T) {
 		},
 		{
 			matcherMap: matcherMap, matcher: matcher,
+			dev:         internal.BlockDevice{Type: string(localv1alpha1.MultiPathType)},
+			spec:        &localv1alpha1.DeviceInclusionSpec{DeviceTypes: []localv1alpha1.DeviceType{"MPATH"}},
+			expectMatch: true, expectErr: false,
+		},
+		{
+			matcherMap: matcherMap, matcher: matcher,
 			dev:         internal.BlockDevice{Type: string(localv1alpha1.Partition)},
 			spec:        &localv1alpha1.DeviceInclusionSpec{DeviceTypes: []localv1alpha1.DeviceType{"DISK", "PART"}},
 			expectMatch: true, expectErr: false,
--- a/diskmaker/discovery/discovery_test.go
+++ b/diskmaker/discovery/discovery_test.go
@@ -151,6 +151,21 @@ func TestIgnoreDevices(t *testing.T) {
 			errMessage: fmt.Errorf("ignored wrong device"),
 		},
 		{
+			label: "don't ignore mpath type",
+			blockDevice: internal.BlockDevice{
+				Name:     "sdc",
+				KName:    "dm-0",
+				ReadOnly: "0",
+				State:    "running",
+				Type:     "mpath",
+			},
+			fakeGlobfunc: func(name string) ([]string, error) {
+				return []string{"removable", "subsytem"}, nil
+			},
+			expected:   false,
+			errMessage: fmt.Errorf("ignored wrong device"),
+		},
+		{
 			label: "ignore read only devices",
 			blockDevice: internal.BlockDevice{
 				Name:     "sdb",
--- a/diskmaker/discovery/discovery_test.go
+++ b/diskmaker/discovery/discovery_test.go
@@ -461,6 +461,50 @@ func TestGetDiscoveredDevices(t *testing.T) {
 				return "/dev/disk/by-id/sda1", nil
 			},
 		},
+		{
+			label: "Case 5: discovering multipath device",
+			blockDevices: []internal.BlockDevice{
+				{
+					Name:       "sda",
+					KName:      "dm-0",
+					FSType:     "xfs",
+					Type:       "mpath",
+					Size:       "62913494528",
+					Model:      "",
+					Vendor:     "",
+					Serial:     "",
+					Rotational: "0",
+					ReadOnly:   "0",
+					Removable:  "0",
+					State:      "running",
+					PartLabel:  "",
+					// We're faking glob function in these test cases and this test would do globbing twice
+					// (first for getting by-id path and second for /dev/mapper path) so we pretend the by-id path is already set.
+					// That way we can test only /dev/mapper globbing and symlink evaluation.
+					PathByID: "/dev/disk/by-id/dm-name-mpatha",
+				},
+			},
+			expected: []v1alpha1.DiscoveredDevice{
+				{
+					DeviceID: "/dev/disk/by-id/dm-name-mpatha",
+					Path:     "/dev/mapper/mpatha",
+					Model:    "",
+					Type:     "mpath",
+					Vendor:   "",
+					Serial:   "",
+					Size:     int64(62913494528),
+					Property: "NonRotational",
+					FSType:   "xfs",
+					Status:   v1alpha1.DeviceStatus{State: v1alpha1.NotAvailable},
+				},
+			},
+			fakeGlobfunc: func(name string) ([]string, error) {
+				return []string{"/dev/mapper/mpatha"}, nil
+			},
+			fakeEvalSymlinkfunc: func(path string) (string, error) {
+				return "/dev/dm-0", nil
+			},
+		},
 	}
 
 	for _, tc := range testcases {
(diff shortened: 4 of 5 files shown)


RomanBednar:
#### Here are my results from manual verification:

Multipath configuration on worker node:
```
sh-4.4# lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINT
loop0         7:0    0   9.8G  0 loop
sda           8:0    0   9.8G  0 disk
`-mpatha    253:0    0   9.8G  0 mpath
sdb           8:16   0   9.8G  0 disk
`-mpatha    253:0    0   9.8G  0 mpath
nvme0n1     259:0    0   120G  0 disk
|-nvme0n1p1 259:1    0     1M  0 part
|-nvme0n1p2 259:2    0   127M  0 part
|-nvme0n1p3 259:3    0   384M  0 part  /boot
`-nvme0n1p4 259:4    0 119.5G  0 part  /sysroot

sh-4.4# multipath -ll
mpatha (333333330000007d1) dm-0 Linux,scsi_debug
size=9.8G features='0' hwhandler='1 alua' wp=rw
`-+- policy='service-time 0' prio=30 status=active
  |- 0:0:0:1 sda 8:0  active ready running
  `- 1:0:0:1 sdb 8:16 active ready running
```

Create LocalVolumeDiscovery in web console and check discovery result:
```
$ oc -n openshift-local-storage get localvolumediscoveryresults.local.storage.openshift.io/discovery-result-ip-10-0-139-122.ec2.internal -o yaml
apiVersion: local.storage.openshift.io/v1alpha1
kind: LocalVolumeDiscoveryResult
metadata:
  creationTimestamp: "2023-04-13T09:33:35Z"
  generation: 1
  labels:
    discovery-result-node: ip-10-0-139-122.ec2.internal
  name: discovery-result-ip-10-0-139-122.ec2.internal
  namespace: openshift-local-storage
  ownerReferences:
  - apiVersion: local.storage.openshift.io/v1alpha1
    kind: LocalVolumeDiscovery
    name: auto-discover-devices
    uid: 17b0f375-0618-46f7-ac04-09cf62a8370d
  resourceVersion: "100723"
  uid: 0e2fe533-0f98-40de-b4a4-9a68eeef7931
spec:
  nodeName: ip-10-0-139-122.ec2.internal
status:
  discoveredDevices:
  - deviceID: ""
    fstype: ""
    model: scsi_debug
    path: /dev/sda
    property: NonRotational
    serial: "2001"
    size: 10485760000
    status:
      state: NotAvailable
    type: disk
    vendor: Linux
  - deviceID: /dev/disk/by-id/dm-name-mpatha
    fstype: ""
    model: ""
    path: /dev/mapper/mpatha
    property: NonRotational
    serial: ""
    size: 10485760000
    status:
      state: NotAvailable
    type: mpath
    vendor: ""
  - deviceID: ""
    fstype: ""
    model: scsi_debug
    path: /dev/sdb
    property: NonRotational
    serial: "2001"
    size: 10485760000
    status:
      state: NotAvailable
    type: disk
    vendor: Linux
  - deviceID: /dev/disk/by-id/dm-name-mpatha
    fstype: ""
    model: ""
    path: /dev/mapper/mpatha
    property: NonRotational
    serial: ""
    size: 10485760000
    status:
      state: NotAvailable
    type: mpath
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol03086c02c6249335e-part1
    fstype: ""
    model: ""
    path: /dev/nvme0n1p1
    property: NonRotational
    serial: ""
    size: 1048576
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol03086c02c6249335e-part2
    fstype: vfat
    model: ""
    path: /dev/nvme0n1p2
    property: NonRotational
    serial: ""
    size: 133169152
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol03086c02c6249335e-part3
    fstype: ext4
    model: ""
    path: /dev/nvme0n1p3
    property: NonRotational
    serial: ""
    size: 402653184
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol03086c02c6249335e-part4
    fstype: xfs
    model: ""
    path: /dev/nvme0n1p4
    property: NonRotational
    serial: ""
    size: 128311082496
    status:
      state: NotAvailable
    type: part
    vendor: ""
  discoveredTimeStamp: "2023-04-13T11:48:54Z"
```

Create LocalVolumeSet in web console and verify:
```
$ oc -n openshift-local-storage get lvset/lvset -o yaml
apiVersion: local.storage.openshift.io/v1alpha1
kind: LocalVolumeSet
metadata:
  creationTimestamp: "2023-04-13T11:19:35Z"
  finalizers:
  - storage.openshift.com/local-volume-protection
  generation: 2
  name: lvset
  namespace: openshift-local-storage
  resourceVersion: "90708"
  uid: b542bc4e-47b7-45c8-b13e-4df607bc0bf7
spec:
  deviceInclusionSpec:
    deviceTypes:
    - mpath
    minSize: 1Gi
  nodeSelector:
    nodeSelectorTerms:
    - matchExpressions:
      - key: kubernetes.io/hostname
        operator: In
        values:
        - ip-10-0-139-122.ec2.internal
  storageClassName: lvset
  volumeMode: Block
status:
  conditions:
  - lastTransitionTime: "2023-04-13T11:19:35Z"
    message: 'DiskMaker: Available'
    status: "True"
    type: DaemonSetsAvailable
  - lastTransitionTime: "2023-04-13T11:19:35Z"
    message: Operator reconciled successfully.
    status: "True"
    type: Available
  observedGeneration: 2
  totalProvisionedDeviceCount: 1
```

Create a PVC:
```
$ oc get pvc/pvc-1 -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
  creationTimestamp: "2023-04-13T11:22:52Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: pvc-1
  namespace: default
  resourceVersion: "91482"
  uid: 0ab38e6b-9e31-4188-aacb-ec030240df3a
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: lvset
  volumeMode: Block
  volumeName: local-pv-d260fe11
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 10000Mi
  phase: Bound
```

Schedule a Pod:
```
apiVersion: v1  
kind: Pod  
metadata:  
  name: example  
  labels:  
    app: httpd  
  namespace: default  
spec:  
  volumes:  
    - name: task-pv-storage  
      persistentVolumeClaim:  
        claimName: pvc-1  
  securityContext:  
    runAsNonRoot: true  
    seccompProfile:  
      type: RuntimeDefault  
  containers:  
    - name: httpd  
      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'  
      ports:  
        - containerPort: 8080  
      volumeDevices:  
        - devicePath: /dev/block  
          name: task-pv-storage  
      securityContext:  
        allowPrivilegeEscalation: false  
        capabilities:  
          drop:  
            - ALL
```

Check pv, pvc and pod status:
```
$ oc get pvc/pvc-1 -o json | jq '.spec.storageClassName'
"lvset"

$ oc get pvc/pvc-1 -o json | jq '.spec.volumeName'
"local-pv-d260fe11"

$ oc get pod/example -o json | jq '.spec.volumes[0].persistentVolumeClaim.claimName'
"pvc-1"

$ oc get pod/example -o json | jq '.status.phase'
"Running"
```

Cross check  block device on worker node with crio:
```
sh-4.4# crictl exec 4bc58ee52214c ls -la /dev/block
brw-rw-rw-. 1 root disk 253, 0 Apr 12 11:46 /dev/block
```

Inspect PersistentVolume and verify volume path:
```
$ oc get pv/local-pv-d260fe11 -o yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/bound-by-controller: "yes"
    pv.kubernetes.io/provisioned-by: local-volume-provisioner-ip-10-0-139-122.ec2.internal
    storage.openshift.com/device-id: dm-name-mpatha
    storage.openshift.com/device-name: dm-0
  creationTimestamp: "2023-04-13T11:20:44Z"
  finalizers:
  - kubernetes.io/pv-protection
  labels:
    kubernetes.io/hostname: ip-10-0-139-122.ec2.internal
    storage.openshift.com/owner-kind: LocalVolumeSet
    storage.openshift.com/owner-name: lvset
    storage.openshift.com/owner-namespace: openshift-local-storage
  name: local-pv-d260fe11
  resourceVersion: "91480"
  uid: 4fd12d94-4530-481b-9a40-fc2adf438660
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 10000Mi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: pvc-1
    namespace: default
    resourceVersion: "91478"
    uid: 0ab38e6b-9e31-4188-aacb-ec030240df3a
  local:
    path: /mnt/local-storage/lvset/dm-name-mpatha
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - ip-10-0-139-122.ec2.internal
  persistentVolumeReclaimPolicy: Delete
  storageClassName: lvset
  volumeMode: Block
status:
  phase: Bound
```

diskmaker/discovery/discovery.go
```diff
@@ -182,8 +182,12 @@ func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.Discove
 			klog.Warningf("failed to parse size for the device %q. Error %v", blockDevice.Name, err)
 		}
 
+		path, err := blockDevice.GetDevPath()
+		if err != nil {
+			klog.Warningf("failed to parse path for the device %q. Error %v", blockDevice.KName, err)
+		}
 		discoveredDevice := v1alpha1.DiscoveredDevice{
-			Path:     fmt.Sprintf("/dev/%s", blockDevice.Name),
+			Path:     path,
```
gnufied, line 190: Can we just get away by changing this line to:

```
Path: fmt.Sprintf("/dev/%s", blockDevice.KName)
```

I am not sure why we need device mapper path from `/dev/mapper` for any of the stuff.

RomanBednar: @gnufied well, we probably don't want to report invalid path in discovery result. That's the only reason I believe. Without the patch you'll get `/dev/mpath` or something like that - such path will never exist for multipath.

gnufied: No, I meant if you used `KName` here, it will point to a valid device. Output of `lsblk` with `KNAME` always points to a valid device.  Having said that, there are pros and cons of both the approaches:

Pro of using `KName` here is - we can use single device mechanism for every device type. 

Pro of using mpath name is, users can easily tell system has selected an mpath device. But then how we show different devices becomes inconsistent. I also think that, mpath device names btw changes between reboots (so can `KName` to be fair), so in the end all of the code must use `dev/disk/by-id` path anyways. But symlink in `/dev/disk/by-id` always points to `KName` device in `/dev/<kname>` location.

RomanBednar: > No, I meant if you used `KName` here, it will point to a valid device. Output of `lsblk` with `KNAME` always points to a valid device. Having said that, there are pros and cons of both the approaches:

Ah, I see. I misread that and though you're suggesting to use what the original code was using, so `blockDevice.Name` for the path.

Those are valid points, the reason I went with `/dev/mapper` was that the original code used `Name` instead of `KName` and I assumed we'd like to keep this naming - volume discovery result looked like this:

```
  - deviceID: /dev/disk/by-id/dm-name-mpatha
    fstype: ""
    model: ""
    path: /dev/mpatha
```

> Pro of using `KName` here is - we can use single device mechanism for every device type.
>
> Pro of using mpath name is, users can easily tell system has selected an mpath device. But then how we show different devices becomes inconsistent. I also think that, mpath device names btw changes between reboots (so can `KName` to be fair), so in the end all of the code must use `dev/disk/by-id` path anyways. But symlink in `/dev/disk/by-id` always points to `KName` device in `/dev/<kname>` location.

Not sure what you mean by "single device mechanism" but I assume it's the way we parse device paths in the code. So with `KName` approach the advantage would be simpler code for sure.

As for the inconsistency in showing the devices: without further refactoring the discovery is not optimal, regardless of which approach of showing path we choose. See the `LocalVolumeDiscoveryResult` I've posted above - we show all underlying devices of mpath (`type: disk`) along with multipath device (`type: mpath`) which is duplicated for every underlying device. Ideally it should show only one mpath? That means adding code that will reliably verify what mpath devices lead to exactly - wondering if it's even worth dealing with.

RomanBednar:
@gnufied  I've updated to code to prevent duplicates in the discovery result and use kname path instead. See the current version is ok. We're expected to backport this to 4.12, I'll clone the Jiras and do the backports once we have a lgtm here.

One thing to note is that the duplicates are now resolved by removing duplicate device IDs, while this works fine it would make much more sense to use `lsblk -M`  once this flag becomes available in our base images.

I've tested the patch again, the discovery result now contains only one device mapper device using KNAME (`/dev/dm-0`).

```
$ oc -n openshift-local-storage get localvolumediscoveryresults.local.storage.openshift.io/discovery-result-ip-10-0-154-173.ec2.internal -o yaml
apiVersion: local.storage.openshift.io/v1alpha1
kind: LocalVolumeDiscoveryResult
metadata:
  creationTimestamp: "2023-04-27T13:20:31Z"
  generation: 1
  labels:
    discovery-result-node: ip-10-0-154-173.ec2.internal
  name: discovery-result-ip-10-0-154-173.ec2.internal
  namespace: openshift-local-storage
  ownerReferences:
  - apiVersion: local.storage.openshift.io/v1alpha1
    kind: LocalVolumeDiscovery
    name: auto-discover-devices
    uid: 8540b196-4c2d-46ea-a364-cb61a6a5adb5
  resourceVersion: "110630"
  uid: 36b8b9f8-a84b-484d-b5a4-7f1a14aef35b
spec:
  nodeName: ip-10-0-154-173.ec2.internal
status:
  discoveredDevices:
  - deviceID: /dev/disk/by-id/scsi-1Linux_scsi_debug_2001
    fstype: ""
    model: scsi_debug
    path: /dev/sda
    property: NonRotational
    serial: "2001"
    size: 10485760000
    status:
      state: NotAvailable
    type: disk
    vendor: Linux
  - deviceID: /dev/disk/by-id/dm-name-mpatha
    fstype: ""
    model: ""
    path: /dev/dm-0
    property: NonRotational
    serial: ""
    size: 10485760000
    status:
      state: Available
    type: mpath
    vendor: ""
  - deviceID: ""
    fstype: ""
    model: scsi_debug
    path: /dev/sdb
    property: NonRotational
    serial: "2001"
    size: 10485760000
    status:
      state: NotAvailable
    type: disk
    vendor: Linux
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0e04d020ab89fcce2-part1
    fstype: ""
    model: ""
    path: /dev/nvme0n1p1
    property: NonRotational
    serial: ""
    size: 1048576
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0e04d020ab89fcce2-part2
    fstype: vfat
    model: ""
    path: /dev/nvme0n1p2
    property: NonRotational
    serial: ""
    size: 133169152
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0e04d020ab89fcce2-part3
    fstype: ext4
    model: ""
    path: /dev/nvme0n1p3
    property: NonRotational
    serial: ""
    size: 402653184
    status:
      state: NotAvailable
    type: part
    vendor: ""
  - deviceID: /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol0e04d020ab89fcce2-part4
    fstype: xfs
    model: ""
    path: /dev/nvme0n1p4
    property: NonRotational
    serial: ""
    size: 128311082496
    status:
      state: NotAvailable
    type: part
    vendor: ""
  discoveredTimeStamp: "2023-04-27T13:20:32Z"
```

diskmaker/discovery/discovery.go
```diff
@@ -197,7 +198,26 @@ func getDiscoverdDevices(blockDevices []internal.BlockDevice) []v1alpha1.Discove
...
+// TODO: remove this and use lsblk with -M flag once base images are updated and lsblk v2.34 or higher is available
+// See: https://github.com/util-linux/util-linux/blob/3be31a106c52e093928afbea2cddbdbe44cfb357/Documentation/releases/v2.34-ReleaseNotes#L18
+func uniqueDevices(sample []v1alpha1.DiscoveredDevice) []v1alpha1.DiscoveredDevice {
+	var unique []v1alpha1.DiscoveredDevice
+	type key struct{ value string }
+	m := make(map[key]int)
+	for _, v := range sample {
+		k := key{v.DeviceID}
+		if i, ok := m[k]; ok {
+			unique[i] = v
```
gnufied, line 214: nit: why did we make a key and type for it? Couldn't we have just used `DeviceId` via something like `map[string]int` ?

gnufied: But another thing is - `deviceID` may not be available always. Could we not have used `v.path` for uniqueness?

RomanBednar: >nit: why did we make a key and type for it? Couldn't we have just used DeviceId via something like map[string]int ?

Sure, but if we would need more criteria for uniqueness instead of just one string using custom type would allow that, like this:

```
...
	type key struct{ value string; value2 string }
	m := make(map[key]int)
	for _, v := range sample {
		k := key{v.DeviceID, v.Path}
...
```

>But another thing is - deviceID may not be available always. Could we not have used v.path for uniqueness?

Correct, `deviceID` can be set to `""` - I missed that. That's even worse with this unique filter because that would hide all devices (except one) that have `deviceID` set to `""`. I think it makes sense to use what I described above and let `uniqueDevices()` use both `v.Path` and `v.DeviceID`?

RomanBednar: >nit: why did we make a key and type for it? Couldn't we have just used DeviceId via something like map[string]int ?

Sure, but if we would need more criteria for uniqueness instead of just one string using custom type would allow that, like this:

```
...
	type key struct{ value string; value2 string }
	m := make(map[key]int)
	for _, v := range sample {
		k := key{v.DeviceID, v.Path}
...
```

>But another thing is - deviceID may not be available always. Could we not have used v.path for uniqueness?

Correct, `deviceID` can be set to `""` - I missed that. That's even worse with this unique filter because that would hide all devices (except one) that have `deviceID` set to `""`. I think it makes sense to use what I described above and let `uniqueDevices()` use both `v.Path` and `v.DeviceID`?

gnufied: should work.

PR #399, opened 2023-04-19
Improve readme

emmahone:
Add high level overview diagram for the operator and for how users are expected to interact with the CRs managed by the operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ graph LR
     User((User)) -->|Creates| LocalVolumeDiscovery
     User -->|Creates| LocalVolumeSet
     User -->|Creates| LocalVolume
-    LocalStorageOperator((LocalStorageOperator)) -->|Manages| LocalVolumeDiscovery
+    LocalStorageOperator((LocalStorageOperator)) -->|Creates/Manages| LocalVolumeDiscoveryResult
     LocalStorageOperator -->|Manages| LocalVolume
     LocalStorageOperator -->|Manages| StorageClass
 ```
--- a/README.md
+++ b/README.md
@@ -4,17 +4,13 @@ Operator for local storage
 ```mermaid
 graph LR
     PVC -->|Requests local storage using a StorageClass| StorageClass
-    StorageClass -->|Defines a LocalVolume| LocalVolume
-    LocalVolume -->|Represents a physical storage device already mounted on a node| Node
+    LocalVolume -->|Represents a physical storage device already mounted/existent on a node| Node
     Pod -->|References a PVC| PVC
     PV -->|References| LocalVolume
-    StorageClass -->|Creates| PV
-    VolumeAttachment -->|Correlates| Node
-    VolumeAttachment -->|Correlates| PV
     LSO((Local Storage Operator)) -->|Manages| LocalVolume
     LSO((Local Storage Operator)) -->|Manages| StorageClass
     LocalVolumeSet -->|Discovers local storage devices on node| Node
-    LocalVolumeSet -->|Automatically manages LocalVolume objects|LocalVolume
+    LocalVolumeSet -->|Creates| PV
 ```
 
 ## Deploying with OLM
@@ -26,7 +22,7 @@ Instructions for using the local storage's must-gather image can be found [here]
 ## How does a user interact with LSO?
 ```mermaid
 graph LR
-    User((User)) -->|Enables| LocalVolumeDiscovery
+    User((User)) -->|Creates| LocalVolumeDiscovery
     User -->|Creates| LocalVolumeSet
     User -->|Creates| LocalVolume
     LocalStorageOperator((LocalStorageOperator)) -->|Manages| LocalVolumeDiscovery
--- a/README.md
+++ b/README.md
@@ -1,8 +1,35 @@
 # local-storage-operator
 Operator for local storage
 
+```mermaid
+graph LR
+    PVC -->|Requests local storage using a StorageClass| StorageClass
+    StorageClass -->|Defines a LocalVolume| LocalVolume
+    LocalVolume -->|Represents a physical storage device already mounted on a node| Node
+    Pod -->|References a PVC| PVC
+    PV -->|References| LocalVolume
+    StorageClass -->|Creates| PV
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,35 @@
...
 ## Deploying with OLM
 Instructions to deploy on OCP >= 4.2 using OLM can be found [here](docs/deploy-with-olm.md)
 
 ## Using the must-gather image with the local storage operator
 Instructions for using the local storage's must-gather image can be found [here](docs/must-gather.md)
+
+## How does a user interact with LSO?
+```mermaid
+graph LR
+    User((User)) -->|Enables| LocalVolumeDiscovery
```
jsafrane, line 29: IMO, user `Creates` LocalVolumeDiscovery.
And LSO creates / manages LocalVolumeDiscoveryResult as the consequence.

README.md
```diff
@@ -1,8 +1,35 @@
...
+```mermaid
+graph LR
+    PVC -->|Requests local storage using a StorageClass| StorageClass
+    StorageClass -->|Defines a LocalVolume| LocalVolume
+    LocalVolume -->|Represents a physical storage device already mounted on a node| Node
+    Pod -->|References a PVC| PVC
+    PV -->|References| LocalVolume
+    StorageClass -->|Creates| PV
+    VolumeAttachment -->|Correlates| Node
+    VolumeAttachment -->|Correlates| PV
```
jsafrane, line 13: LSO PVs are not "attachable" and do not use VolumeAttachments.

README.md
```diff
@@ -1,8 +1,35 @@
...
 Operator for local storage
 
+```mermaid
+graph LR
+    PVC -->|Requests local storage using a StorageClass| StorageClass
+    StorageClass -->|Defines a LocalVolume| LocalVolume
+    LocalVolume -->|Represents a physical storage device already mounted on a node| Node
+    Pod -->|References a PVC| PVC
+    PV -->|References| LocalVolume
+    StorageClass -->|Creates| PV
```
jsafrane, line 11: No, LSO creates PVs. StorageClass is just for reference and is not really used anywhere in PV creation / deletion / binding.

README.md
```diff
@@ -1,8 +1,35 @@
 # local-storage-operator
 Operator for local storage
 
+```mermaid
+graph LR
+    PVC -->|Requests local storage using a StorageClass| StorageClass
+    StorageClass -->|Defines a LocalVolume| LocalVolume
```
jsafrane, line 7: StorageClass is just for reference and is not really used anywhere in PV creation / deletion / binding. The don't define anything.

README.md
```diff
@@ -1,8 +1,35 @@
...
+    LocalVolume -->|Represents a physical storage device already mounted on a node| Node
```
jsafrane, line 8: LocalVolume does not need to be mounted on a node, it can be just a block device.

README.md
```diff
@@ -1,8 +1,35 @@
...
+    Pod -->|References a PVC| PVC
+    PV -->|References| LocalVolume
+    StorageClass -->|Creates| PV
+    VolumeAttachment -->|Correlates| Node
+    VolumeAttachment -->|Correlates| PV
+    LSO((Local Storage Operator)) -->|Manages| LocalVolume
+    LSO((Local Storage Operator)) -->|Manages| StorageClass
+    LocalVolumeSet -->|Discovers local storage devices on node| Node
+    LocalVolumeSet -->|Automatically manages LocalVolume objects|LocalVolume
```
jsafrane, line 17: LocalVolumeSet does not manages LocalVolumes, it directly creates PVs.

README.md
emmahone: Thanks for reviewing/clarifying. This should be fixed in https://github.com/openshift/local-storage-operator/commit/f8acb21ec74b51b9fdfde6ab9f8c0a8d71ad3ea9

README.md
emmahone: Do you have a recommendation about what we should state as the relationship between the localvolume and the node? Right now I just have "Represents a physical storage device already mounted/existent on a node".

README.md
emmahone: Thanks for catching that. Changed the flow so that LSO creates the PVs in https://github.com/openshift/local-storage-operator/commit/f8acb21ec74b51b9fdfde6ab9f8c0a8d71ad3ea9

README.md
emmahone: Changed flow so that LVS creates PVs https://github.com/openshift/local-storage-operator/commit/f8acb21ec74b51b9fdfde6ab9f8c0a8d71ad3ea9#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R13

README.md
emmahone: Agreed. I made this change and added LVDR as well in https://github.com/openshift/local-storage-operator/commit/f8acb21ec74b51b9fdfde6ab9f8c0a8d71ad3ea9#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R25

emmahone:
Thanks so much for the review @jsafrane ! I think I have addressed most of those points you made. Is there a best practice for contributing to LSO? Should I squash all these changes into a single commit?

README.md
```diff
@@ -1,8 +1,31 @@
...
+
+## How does a user interact with LSO?
+```mermaid
+graph LR
+    User((User)) -->|Creates| LocalVolumeDiscovery
+    User -->|Creates| LocalVolumeSet
+    User -->|Creates| LocalVolume
+    LocalStorageOperator((LocalStorageOperator)) -->|Creates/Manages| LocalVolumeDiscoveryResult
+    LocalStorageOperator -->|Manages| LocalVolume
+    LocalStorageOperator -->|Manages| StorageClass
```
gnufied, line 30: If we are going to say local-storage-operator manages `LocalVolume` objects then it also manages `LocalVolumeSet` objects. But then may be neither of them is true, because LSO only reconciles these objects. It is user which creates and deletes `LocalVolume` and `LocalVolumeSet` objects.

PR #409, opened 2023-05-19
Fix cli instructions

gnufied:


gnufied:
cc @openshift/storage

PR #410, opened 2023-06-09
refactor sync_bundle script

RomanBednar:
The idea is to make sync_bundle script more robust, easier to use and less susceptible to user errors when called.

Here are some specific issues found when working with old `sync_bundle` (ruby) that are being addressed by this change:

- positional `bundle` argument checking
	* script does not warn that the positional argument `bundle` was not used
	* without it the script won't produce bundle and index images (even if  `-i` and `-b` flags are used) and no warning is printed
	* `bundle` positional argument is missing in help message
* no dependency checks
	* exit without a warning if `opm` command is not installed 
	* return code is `0` which hides the issue even more
- unhelpful warnings about Containerfiles
	- script checks Containerfiles content
		- one condition for the warning is usage of `yum` in Containerfile
		- second condition is `registry` string presence in Container file (line with `FROM` directive) which should indicate private registry
	- these warnings lack explanation and can be removed

See [HACKING.md](https://github.com/openshift/local-storage-operator/pull/410/files#diff-9769a3f5504af3cd735e08f078b94f827bdd51eacac2e9d80ac9761bc485171fR47-R53) for usage.


gnufied:
It looks good to me. I am not entirely sold on using Go for scripting but I also like more simpler approach we took for efs - https://github.com/openshift/aws-efs-csi-driver-operator/blob/master/hack/create-bundle and gcp - https://github.com/openshift/gcp-filestore-csi-driver-operator/blob/main/hack/create-bundle

RomanBednar:
Closing in favor of: https://github.com/openshift/local-storage-operator/pull/433

PR #411, opened 2023-06-27
OCPBUGS-15358: Improve logging existing symlinks

gnufied:
Fixes https://issues.redhat.com/browse/OCPBUGS-15358

After the fix:

```
E0627 18:20:18.432300  423801 reconcile.go:371] "skipping provisioning of PV" err="found existing symlinks for device sdb in /mnt/local-storage"
E0627 18:20:18.432317  423801 reconcile.go:170] provisioning failed for sdb: found existing symlinks for device sdb in /mnt/local-storage
2023-06-27T18:20:18.432Z        ERROR   controller/controller.go:329    Reconciler error        {"controller": "localvolumeset", "controllerGroup": "local.storage.openshift.io", "controllerKind": "LocalVolumeSet", "LocalVolumeSet": {"name":"example","name
space":"openshift-local-storage"}, "namespace": "openshift-local-storage", "name": "example", "reconcileID": "cb41a2ae-d354-409b-a87b-afbc7dc39239", "error": "could not provision disk: found existing symlinks for device sdb in /mnt/local-storage"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
```

### DIFF
--- a/diskmaker/controllers/lvset/reconcile.go
+++ b/diskmaker/controllers/lvset/reconcile.go
@@ -346,7 +346,7 @@ func (r *LocalVolumeSetReconciler) provisionPV(
 	unlockFunc := func() {
 		err := pvLock.Unlock()
 		if err != nil {
-			klog.ErrorS(err, "failed to unlock device")
+			klog.ErrorS(err, "failed to unlock device", "disk", devLabelPath)
 		}
 	}
 	defer unlockFunc()
@@ -369,9 +369,9 @@ func (r *LocalVolumeSetReconciler) provisionPV(
 		}
 		err := fmt.Errorf("found existing symlinks for device %s in %s", dev.KName, filepath.Dir(symLinkDir))
 		klog.ErrorS(err, "skipping provisioning of PV")
-		return err
+		return nil
 	} else if err != nil || !pvLocked { // locking failed for some other reasion
-		klog.ErrorS(err, "not provisioning, could not get lock")
+		klog.ErrorS(err, "not provisioning, could not get lock", "disk", devLabelPath)
 		return err
 	}
 
--- a/diskmaker/controllers/lvset/reconcile.go
+++ b/diskmaker/controllers/lvset/reconcile.go
@@ -164,11 +164,10 @@ func (r *LocalVolumeSetReconciler) Reconcile(ctx context.Context, request ctrl.R
 		r.eventReporter.Report(lvset, newDiskEvent(diskmaker.FoundMatchingDisk, "provisioning matching disk", blockDevice.KName, corev1.EventTypeNormal))
 		err = r.provisionPV(lvset, blockDevice, *storageClass, mountPointMap, symlinkSourcePath, symlinkPath, idExists)
 		if err != nil {
-			msg := fmt.Sprintf("provisioning failed for %s: %v",
-				blockDevice.Name, err)
+			msg := fmt.Sprintf("provisioning failed for %s: %v", blockDevice.Name, err)
 			r.eventReporter.Report(lvset, newDiskEvent(diskmaker.ErrorProvisioningDisk, msg, blockDevice.KName, corev1.EventTypeWarning))
 			klog.Error(msg)
-			return ctrl.Result{}, fmt.Errorf("could not provision disk: %w", err)
+			continue
 		}
 
 		klog.InfoS("provisioning succeeded", "blockDevice", blockDevice.Name)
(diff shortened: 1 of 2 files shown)


diskmaker/controllers/lvset/reconcile.go
```diff
@@ -369,9 +369,9 @@ func (r *LocalVolumeSetReconciler) provisionPV(
 		}
 		err := fmt.Errorf("found existing symlinks for device %s in %s", dev.KName, filepath.Dir(symLinkDir))
 		klog.ErrorS(err, "skipping provisioning of PV")
-		return err
+		return nil
```
mpatlasov, line 372: This looks as a typo. You probably meant `return err` as it was introduced in a previous commit (https://github.com/openshift/local-storage-operator/pull/411/commits/e9102ca1a3a1d1f8ad03f8b36f1b63be2f4f25ed#diff-359de61f56f71878050948f3339fb3d1344c31038eaa1f87072efc729ffd6edbR372)

gnufied: No that is intentional unfortunately. So if we return error from here, lets say there are 5 discovered disks on the node, then it will stop provisioning of all disks and stop. But if we log error and carry on, then it will at least provision rest 4 of them. So that is why e2e was failing.

gnufied: So I pushed a change, which returns error from here and does `continue` from the caller, so as next disk still gets provisioned. It would be a behaviour change of LSO I think, but I don't see particular downsides.

mpatlasov: Please update output snippet in PR description because it (currently) assumes `return err`:
```
E0627 18:20:18.432300  423801 reconcile.go:371] "skipping provisioning of PV" err="found existing symlinks for device sdb in /mnt/local-storage"
E0627 18:20:18.432317  423801 reconcile.go:170] provisioning failed for sdb: found existing symlinks for device sdb in /mnt/local-storage
2023-06-27T18:20:18.432Z        ERROR   controller/controller.go:329    Reconciler error        {"controller": "localvolumeset", "controllerGroup": "local.storage.openshift.io", "controllerKind": "LocalVolumeSet", "LocalVolumeSet": {"name":"example","name
space":"openshift-local-storage"}, "namespace": "openshift-local-storage", "name": "example", "reconcileID": "cb41a2ae-d354-409b-a87b-afbc7dc39239", "error": "could not provision disk: found existing symlinks for device sdb in /mnt/local-storage"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
```

gnufied: I changed the code to return error here. It will log an event and process error now.

mpatlasov: I understand your point about stop provisioning if error, but resulting code looks too ugly and hard to understand. The caller of Reconcile() has the following clause [here](https://github.com/openshift/local-storage-operator/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L341-L345):
```
	default:
		// Finally, if no error occurs we Forget this item so it does not
		// get queued again until another change happens.
		c.Queue.Forget(obj)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelSuccess).Inc()
```
This made me think that if we return `nil` despite of error, the Reconcile() won't be called again, like "everything is fine, we have reconciled everything, no more efforts needed". So, even if an admin manually cleans-up symlink, we won't catch it.

Would it be worthy to rework the Reconcile() a little bit to keep looping over other disks even if an error detected? Then, in the end of function, we'd return either `nil` or `err` (if any error detected). In case of more than one error, we can return the first or the last one - it doesn't matter, imho.

gnufied: No, we call `Reconcile` every 1 minute regardless of error or not error during reconciliation. See - https://github.com/openshift/local-storage-operator/blob/547bf1d2f1c6aec8812475911cbd9efe585fb758/diskmaker/controllers/lvset/reconcile.go#L205

So if admin removes the symlink, the device should be reconciled again within a minute and PV should be provisioned.

mpatlasov: My comment immediately above was about previous version of PR. While I was typing it, new version came, and that new comment:

> I changed the code to return error here. It will log an event and process error now.

Now, only one question remains: is it OK that the [caller](https://github.com/openshift/local-storage-operator/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L323) won't see that error we reported?

PR #413, opened 2023-07-24
OCPBUGS-16595: Add nodeSelector to sample LocalVolume CR

jsafrane:
`nodeSelector` should be present in majority of LocalVolume, so add it to the example.


jsafrane:
cc @openshift/storage

PR #417, opened 2023-08-02
OCPBUGS-16745: prefer wwn, scsi and nvme devices

gnufied:
Prefer wwn, scsi and nvme device symlinks first.

Fixes https://issues.redhat.com/browse/OCPBUGS-16745

Note fore reviewers - I know this is slower than processing `/dev/disk/by-id` exactly once but I have benchmarked and evaluated various approaches.

One of the more faster solution is using a prefixtree (or a trie):

```
preferredPatterns := []string{"wwn", "scsi", "nvme",  ""}
	for _, pattern := range preferredPatterns {
		preferrdPaths := tree.FindKeys(pattern)
		for _, path := range preferrdPaths {
			matchPath, err := matchSymlink(path)
			if err != nil {
				return "", err
			}
			if matchPath != "" {
				return matchPath, nil
			}
		}
	}
```

I was using "github.com/beevik/prefixtree" library for that and is generally 30-40% faster. Not sure how much it matters in actual practice though.

### DIFF
--- a/internal/diskutil.go
+++ b/internal/diskutil.go
@@ -126,7 +126,7 @@ func (b BlockDevice) HasChildren() (bool, error) {
 
 // HasBindMounts checks for bind mounts and returns mount point for a device by parsing `proc/1/mountinfo`.
 // HostPID should be set to true inside the POD spec to get details of host's mount points inside `proc/1/mountinfo`.
-func (b BlockDevice) HasBindMounts() (bool, string, error) {
+func (b *BlockDevice) HasBindMounts() (bool, string, error) {
 	data, err := ioutil.ReadFile(mountFile)
 	if err != nil {
 		return false, "", fmt.Errorf("failed to read file %s: %v", mountFile, err)
@@ -161,7 +161,7 @@ func (b BlockDevice) GetDevPath() (path string, err error) {
 }
 
 // GetPathByID check on BlockDevice
-func (b BlockDevice) GetPathByID() (string, error) {
+func (b *BlockDevice) GetPathByID() (string, error) {
 
 	// return if previously populated value is valid
 	if len(b.PathByID) > 0 && strings.HasPrefix(b.PathByID, DiskByIDDir) {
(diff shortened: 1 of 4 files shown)


internal/diskutil.go
```diff
@@ -171,12 +173,17 @@ func (b BlockDevice) GetPathByID() (string, error) {
...
 	}
 	b.PathByID = ""
-	diskByIDDir := filepath.Join(DiskByIDDir, "/*")
-	paths, err := FilePathGlob(diskByIDDir)
-	if err != nil {
-		return "", fmt.Errorf("could not list files in %q: %w", DiskByIDDir, err)
+	allDisks := []string{}
+	// We are going to prefer wwn, scsi and then nvme- disk-ids in that order
+	for _, pattern := range preferredPatterns {
+		disks, err := fileGlobFunc(pattern, DiskByIDDir)
```
jsafrane, line 179: THB, I don't like calling `FilePathGlob` that often.

I would personally get all by-id disks as today + sort them by preferred prefix (i.e. `/dev/disk/by-id/wwn-` would be before `/dev/disk/by-id/google-`) and _then_ filter them below using `PathEvalsToDiskLabel`. What it wwn-* symlink does not lead to `/dev/$KNAME`?

gnufied: I benchmarked that option - https://gist.github.com/gnufied/47f7d580d953b98178772eddc0cbfcb5 . Using glob is faster than sorting and filtering. See - https://gist.github.com/gnufied/47f7d580d953b98178772eddc0cbfcb5#file-main-go-L24

gnufied: yeah if wwn- doesn't lead to /dev/$KNAME, then we will have to move to scsi- and then to nvme- and finally to other names.  Fastest option is clearly using the trie.

gnufied: Sorry I ended up using grouped by prefix approach in the end.  The bottleneck turned out to be regex matching and using a simpler `strings.HasPrefix` results in much faster code.

internal/diskutil.go
```diff
@@ -171,21 +171,47 @@ func (b BlockDevice) GetPathByID() (string, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("could not list files in %q: %w", DiskByIDDir, err)
-	}
-	for _, path := range paths {
-		isMatch, err := PathEvalsToDiskLabel(path, b.KName)
-		if err != nil {
-			return "", err
+		return "", fmt.Errorf("error listing files in %s: %v", DiskByIDDir, err)
+	}
+	preferredPatterns := []string{"wwn", "scsi", "nvme"}
```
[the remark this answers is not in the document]
gnufied: fixed.

internal/diskutil.go
```diff
@@ -171,21 +171,47 @@ func (b BlockDevice) GetPathByID() (string, error) {
...
+
+	// sortedSymlinks sorts symlinks in 4 buckets.
+	// 	- [0] - syminks that match wwn
+	//	- [1] - symlinks that match scsi
+	//	- [2] - symlinks that match nvme
+	//	- [3] - symlinks that does not any of these
+	sortedSymlinks := [4][]string{}
```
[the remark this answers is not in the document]
gnufied: fixed.

PR #420, opened 2023-08-09
remove unnecessary builder image from Dockerfile.mustgather

dobsonj:
Follow up PR from https://github.com/openshift/secrets-store-csi-driver-operator/pull/9#discussion_r1288132661

The builder image is not really necessary for the must-gather image.

/cc @openshift/storage @jsafrane


dobsonj:
There is no user-visible change, it just simplifies how we build the must-gather image
/label docs-approved
/label px-approved
/label qe-approved

PR #431, opened 2023-09-27
STOR-1377: Add option `forceWipeDevicesAndDestroyAllData` to wipe local volumes

mpatlasov:
An exampe of LocalVolume with new option set:
```
apiVersion: "local.storage.openshift.io/v1"
kind: "LocalVolume"
metadata:
  name: "local-disks"
  namespace: "openshift-local-storage"
spec:
  nodeSelector:
    nodeSelectorTerms:
    - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - ip-10-0-7-58.us-west-2.compute.internal
  storageClassDevices:
    - storageClassName: "local-sc"
      forceWipeDevicesAndDestroyAllData: true
      volumeMode: Filesystem
      fsType: xfs
      devicePaths:
        - /dev/loop0
```

`forceWipeDevicesAndDestroyAllData` is optional. The default is false. If it is set explicitly to true (as in example above), corresponding `devicePaths` will be wiped out before exposing as PVs. Wiping must happen only once, along with creating symlink in `/mnt/local-storage/<storageClassName>/`.

### DIFF
--- a/api/v1/localvolume_types.go
+++ b/api/v1/localvolume_types.go
@@ -64,6 +64,9 @@ type StorageClassDevice struct {
 	// A list of device paths which would be chosen for local storage.
 	// For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]
 	DevicePaths []string `json:"devicePaths,omitempty"`
+	// This option will destroy all leftover data on the devices before they're used as PersistentVolumes. Use with care.
+	// +optional
+	ForceWipeDevicesAndDestroyAllData bool `json:"forceWipeDevicesAndDestroyAllData,omitempty"`
 }
 
 // LocalVolumeStatus defines the observed state of LocalVolume
--- a/diskmaker/controllers/lv/reconcile.go
+++ b/diskmaker/controllers/lv/reconcile.go
@@ -64,6 +64,7 @@ type DiskLocation struct {
 	userProvidedPath string
 	diskID           string
 	blockDevice      internal.BlockDevice
+	forceWipe        bool
 }
 
 type LocalVolumeReconciler struct {
@@ -150,6 +151,17 @@ func (r *LocalVolumeReconciler) createSymlink(
 		return false
 	}
 
+	if deviceNameLocation.forceWipe {
+		cmd := exec.Command("wipefs", "-a", "-f", symLinkSource)
+		err = cmd.Run()
+		if err != nil {
+			msg := fmt.Sprintf("error wipefs device %s: %v", symLinkSource, err)
+			r.eventSync.Report(r.localVolume, newDiskEvent(ErrorCreatingSymLink, msg, symLinkSource, corev1.EventTypeWarning))
+			klog.Error(msg)
+			return false
+		}
+	}
+
 	err = os.Symlink(symLinkSource, symLinkTarget)
 	if err != nil {
 		msg := fmt.Sprintf("error creating symlink %s: %v", symLinkTarget, err)
@@ -211,6 +223,7 @@ func (r *LocalVolumeReconciler) generateConfig() *DiskConfig {
 	storageClassDevices := r.localVolume.Spec.StorageClassDevices
 	for _, storageClassDevice := range storageClassDevices {
 		disks := new(Disks)
+		disks.ForceWipeDevicesAndDestroyAllData = storageClassDevice.ForceWipeDevicesAndDestroyAllData
 		if len(storageClassDevice.DevicePaths) > 0 {
 			disks.DevicePaths = storageClassDevice.DevicePaths
 		}
@@ -486,17 +499,18 @@ func (r *LocalVolumeReconciler) findMatchingDisks(diskConfig *DiskConfig, blockD
 
 	// blockDeviceMap is a map of storageclass and device locations
 	blockDeviceMap := make(map[string][]DiskLocation)
-	addDiskToMap := func(scName, stableDeviceID, diskName, userDevicePath string, blockDevice internal.BlockDevice) {
+	addDiskToMap := func(scName, stableDeviceID, diskName, userDevicePath string, blockDevice internal.BlockDevice, forceWipe bool) {
 		deviceArray, ok := blockDeviceMap[scName]
 		if !ok {
 			deviceArray = []DiskLocation{}
 		}
-		deviceArray = append(deviceArray, DiskLocation{diskName, userDevicePath, stableDeviceID, blockDevice})
+		deviceArray = append(deviceArray, DiskLocation{diskName, userDevicePath, stableDeviceID, blockDevice, forceWipe})
 		blockDeviceMap[scName] = deviceArray
 	}
 
 	for storageClass, disks := range diskConfig.Disks {
 		devicePaths := disks.DevicePaths
+		forceWipe := disks.ForceWipeDevicesAndDestroyAllData
 		for _, devicePath := range devicePaths {
 			// handle user provided device_ids first
 			if strings.HasPrefix(devicePath, diskByIDPrefix) {
@@ -511,7 +525,7 @@ func (r *LocalVolumeReconciler) findMatchingDisks(diskConfig *DiskConfig, blockD
 				// We need to make sure that requested device is not already mounted.
 				blockDevice, matched := hasExactDisk(blockDevices, baseDeviceName)
 				if matched {
-					addDiskToMap(storageClass, matchedDeviceID, matchedDiskName, devicePath, blockDevice)
+					addDiskToMap(storageClass, matchedDeviceID, matchedDiskName, devicePath, blockDevice, forceWipe)
 				}
 			} else {
 				// handle anything other than device ids here - such as:
@@ -535,10 +549,10 @@ func (r *LocalVolumeReconciler) findMatchingDisks(diskConfig *DiskConfig, blockD
 					if err != nil {
 						klog.ErrorS(err, "unable to find disk ID for local pool",
 							"diskName", diskDevPath)
-						addDiskToMap(storageClass, "", diskDevPath, devicePath, blockDevice)
+						addDiskToMap(storageClass, "", diskDevPath, devicePath, blockDevice, forceWipe)
 						continue
 					}
-					addDiskToMap(storageClass, matchedDeviceID, diskDevPath, devicePath, blockDevice)
+					addDiskToMap(storageClass, matchedDeviceID, diskDevPath, devicePath, blockDevice, forceWipe)
 					continue
 				} else {
 					r.logDeviceError(diskDevPath)
--- a/diskmaker/controllers/lv/reconcile_test.go
+++ b/diskmaker/controllers/lv/reconcile_test.go
@@ -206,7 +206,7 @@ func TestCreateSymLinkByDeviceID(t *testing.T) {
 	}
 	d, _ := getFakeDiskMaker(t, tmpSymLinkTargetDir, lv, sc)
 	d.fsInterface = FakeFileSystemInterface{}
-	diskLocation := DiskLocation{fakeDisk.Name(), fakeDiskByID.Name(), fakeDisk.Name(), internal.BlockDevice{}}
+	diskLocation := DiskLocation{fakeDisk.Name(), fakeDiskByID.Name(), fakeDisk.Name(), internal.BlockDevice{}, false}
 
 	d.runtimeConfig = &provCommon.RuntimeConfig{
 		UserConfig: &provCommon.UserConfig{
@@ -253,7 +253,7 @@ func TestCreateSymLinkByDeviceName(t *testing.T) {
 
 	d, _ := getFakeDiskMaker(t, tmpSymLinkTargetDir, lv, sc)
 	d.fsInterface = FakeFileSystemInterface{}
-	diskLocation := DiskLocation{fakeDisk.Name(), "", fakeDisk.Name(), internal.BlockDevice{}}
+	diskLocation := DiskLocation{fakeDisk.Name(), "", fakeDisk.Name(), internal.BlockDevice{}, false}
 	d.createSymlink(diskLocation, fakeDisk.Name(), path.Join(tmpSymLinkTargetDir, "diskName"), false)
 
 	// assert that target symlink is created for disk name when no disk ID is available
--- a/diskmaker/controllers/lv/reconcile_test.go
+++ b/diskmaker/controllers/lv/reconcile_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/local-storage-operator/internal"
 
 	localv1 "github.com/openshift/local-storage-operator/api/v1"
+	"github.com/openshift/local-storage-operator/test-framework/testutils"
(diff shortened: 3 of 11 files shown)


api/v1/localvolume_types.go
```diff
@@ -64,6 +64,9 @@ type StorageClassDevice struct {
 	// A list of device paths which would be chosen for local storage.
 	// For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]
 	DevicePaths []string `json:"devicePaths,omitempty"`
+	// Boolean flag to enforce wiping devices before first use
```
dobsonj, line 67: Maybe be more explicit what we mean by "wiping devices", something like "wiping existing filesystem signatures from the device".

diskmaker/controllers/lv/reconcile.go
```diff
@@ -87,11 +88,11 @@ func (r *LocalVolumeReconciler) createSymlink(
 	symLinkSource string,
 	symLinkTarget string,
 	idExists bool,
-) bool {
+) (bool, bool) {
```
dobsonj, line 91: Can you change these bool return values to a single return value using an enum? I wasn't sure what each one meant until I looked where this function was called: `shouldCreatePV, symlinkCreated`.

There are only 3 combinations that this function returns, so something like this could work:
```
type SymlinkStatus int

const (
    NotCreated SymlinkStatus = 0 // false, false
    SymlinkExists SymlinkStatus = 1 // true, false
    SymlinkCreated SymlinkStatus = 2 // true, true
)
```

api/v1/localvolume_types.go
```diff
@@ -64,6 +64,9 @@ type StorageClassDevice struct {
 	// A list of device paths which would be chosen for local storage.
 	// For example - ["/dev/sda", "/dev/sdb", "/dev/disk/by-id/ata-crucial"]
 	DevicePaths []string `json:"devicePaths,omitempty"`
+	// Boolean flag to enforce wiping existing filesystem signatures from the devices before first use.
```
jsafrane, line 67: I would add something more explicit, like "This option will destroy all leftover data on the devices before they're used as PersistentVolumes. Use with care."
It should sound dangerous.

diskmaker/controllers/lv/reconcile.go
```diff
@@ -415,6 +426,16 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 					common.LocalVolumeOwnerNamespaceForPV: r.localVolume.Namespace,
 				}
 
+				if createSymlinkStatus == SymlinkCreated && deviceNameLocation.forceWipe {
```
jsafrane, line 429: What if a diskmaker process is restarted after `r.createSymlink` above and before this line? The previous diskmaker created the symlink, but did not have time to wipe the volume, and the new one found the existing symlink and did not wipe the volume either, because it got `SymlinkExists`.

IMO, diskmaker should wipe volumes *before* creating the symlink to have it bulletproof. In the worst case, it will be wiped several times, but it will be wiped before PV is created.

mpatlasov: Good catch, fully agreed!

diskmaker/controllers/lv/reconcile.go
```diff
@@ -415,6 +426,16 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+					cmd := exec.Command("/usr/sbin/wipefs", "-a", "-f", source)
+					err = cmd.Run()
```
gnufied, line 431: nit: is it necessary to hardcode location of `wipefs` ?

mpatlasov:
The failure of `ci/prow/e2e-operator-extended` is well-known problem (not caused by this PR), let's ignore it for now.

diskmaker/controllers/lv/reconcile_test.go
```diff
@@ -253,7 +253,7 @@ func TestCreateSymLinkByDeviceName(t *testing.T) {
 
 	d, _ := getFakeDiskMaker(t, tmpSymLinkTargetDir, lv, sc)
 	d.fsInterface = FakeFileSystemInterface{}
-	diskLocation := DiskLocation{fakeDisk.Name(), "", fakeDisk.Name(), internal.BlockDevice{}}
+	diskLocation := DiskLocation{fakeDisk.Name(), "", fakeDisk.Name(), internal.BlockDevice{}, false}
```
jsafrane, line 256: I think it would be useful to add a linux-only e2e test that actually provides a small few MB ext2 or fat filesystem and expects it to be wiped after `createSymlink`.

For example, this test calls `createTmpFile` to create a file that the test uses as a "device", it should be possible either to copy a small image of a filesystem there or even create an empty 1MiB file and call `mkfs.ext2` on that file + expect it to be wiped (i.e. `file -s` does not return `ext2`)

mpatlasov:
Last push makes some cleanup in `reconcile_test.go` and then adds new unit test that extends existing `TestCreateSymLinkByDeviceName` adding checks for `forceWipeDevicesAndDestroyAllData`.

diskmaker/controllers/lv/reconcile_test.go
```diff
@@ -318,34 +347,51 @@ func getDeiveIDs() []string {
...
 		}
 	}
 	return false
 }
 
+func formatAsExt4(t *testing.T, fname string) {
+	cmd := exec.Command("mkfs.ext4", fname)
+	err := cmd.Run()
+	if err != nil {
+		t.Fatalf("error formatting file: %v", fname)
```
RomanBednar, line 386: This would hide the actual command error if any. It's better to show it with something like:

`t.Fatalf("error formatting file: %v Command error: %v", fname, cmd.Err)`

diskmaker/controllers/lv/reconcile_test.go
```diff
@@ -318,34 +347,51 @@ func getDeiveIDs() []string {
...
-		if strings.Contains(f.Name(), file) {
+	for _, d := range dentries {
+		if strings.Contains(d.Name(), file) {
 			return true
 		}
 	}
 	return false
 }
 
+func formatAsExt4(t *testing.T, fname string) {
```
RomanBednar, line 382: I believe this should move under [VolumeUtil interface](https://github.com/openshift/local-storage-operator/blob/4e05a300cc0a5ce7f2755b4908332125bcf5074f/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/util/volume_util.go#L24) and since this is Linux specific the implementation should be in [this](https://github.com/openshift/local-storage-operator/blob/4e05a300cc0a5ce7f2755b4908332125bcf5074f/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/util/volume_util_linux.go) file.

Then the implementation in `volume_util_unsupported.go` should get these new methods (`formatAsExt4` and `hasExt4`) as well to throw unsupported error when your test runs on different platform.

diskmaker/controllers/lv/reconcile_test.go
```diff
@@ -318,34 +347,51 @@ func getDeiveIDs() []string {
...
+	cmd := exec.Command("mkfs.ext4", fname)
+	err := cmd.Run()
+	if err != nil {
+		t.Fatalf("error formatting file: %v", fname)
+	}
+}
+
+func hasExt4(fname string) bool {
```
RomanBednar, line 390: Same here, see comment for `formatAsExt4`.

diskmaker/controllers/lv/reconcile_test.go
mpatlasov: @RomanBednar , [VolumeUtil interface](https://github.com/openshift/local-storage-operator/blob/4e05a300cc0a5ce7f2755b4908332125bcf5074f/vendor/sigs.k8s.io/sig-storage-local-static-provisioner/pkg/util/volume_util.go#L24) comes from upstream `sig-storage-local-static-provisioner` , and it doesn't need `formatAsExt4()`.

I don't think we can (or have to) pollute upstream repo with some code that is required only for LSO, especially only for LSO tests. Theoretically, we could implement similar stuff (`volume_util*.go`) somewhere downstream, but before doing that -- Do we support LSO on any different (than linux) platform? If not, why we're intersted about running tests on different platforms?

mpatlasov:
Last push fixes `t.Fatalf(...)`  message in `formatAsExt4()` per [this comment](https://github.com/openshift/local-storage-operator/pull/431#pullrequestreview-1664024460)

diskmaker/controllers/lv/reconcile_test.go
mpatlasov: @RomanBednar , are you OK with this simple implementation (`return cmd.Run() == nil`) or it's worthy to check ExitCode and print explicit error message if we fail for another reason, e.g. `fsck.ext4` is not in PATH? (I intentionally preferred code simplicity there)

RomanBednar: @mpatlasov It's the same case as with `formatAsExt4` - let's say the `fsck.ext4` command gets called with wrong path by mistake (for `-n` arg), it will just return false and will be difficult to spot and debug. Not a hard opinion on that one, but generally it's better not to hide the errors.

diskmaker/controllers/lv/reconcile_test.go
RomanBednar: @mpatlasov Ah, it's for testing only. So you're right, it should not be in `sig-storage-local-static-provisioner`. 

I don't think we support anything else than Linux in LSO, but the unit test itself will fail weird if not run on Linux:

```
=== RUN   TestWipeDeviceWhenCreateSymLinkByDeviceName
=== RUN   TestWipeDeviceWhenCreateSymLinkByDeviceName/forceWipeDevicesAndDestroyAllData_is_False
    reconcile_test.go:386: error formatting file: /var/folders/98/w1dp6fq94k53hgy7ml6xqz8h0000gn/T/diskName4038852681; command error: exec: "mkfs.ext4": executable file not found in $PATH
=== RUN   TestWipeDeviceWhenCreateSymLinkByDeviceName/forceWipeDevicesAndDestroyAllData_is_True
    reconcile_test.go:386: error formatting file: /var/folders/98/w1dp6fq94k53hgy7ml6xqz8h0000gn/T/diskName3925205785; command error: exec: "mkfs.ext4": executable file not found in $PATH
--- FAIL: TestWipeDeviceWhenCreateSymLinkByDeviceName (0.00s)
    --- FAIL: TestWipeDeviceWhenCreateSymLinkByDeviceName/forceWipeDevicesAndDestroyAllData_is_False (0.00s)
```

If we do something similar to `VolumeUtil` like [this](https://github.com/RomanBednar/local-storage-operator/commit/227a51d407850441d622d6e2f8b691216691f0cd) we can directly say it's unsupported at build:

```
--- FAIL: TestWipeDeviceWhenCreateSymLinkByDeviceName (0.00s)
    --- FAIL: TestWipeDeviceWhenCreateSymLinkByDeviceName/forceWipeDevicesAndDestroyAllData_is_False (0.00s)
panic: FormatAsExt4 is unsupported in this build [recovered]
	panic: FormatAsExt4 is unsupported in this build
```

The ideal case I believe would be to change `make test` to run tests in a container, but that's out of scope of this PR.

Maybe the platform specific builds are not worth the hassle in this case, feel free to keep it as it is for now.

mpatlasov: @RomanBednar thank you for example implementation of `test-framework/testutils/volume_util*.go` -- that's what I meant saying we could implement similar stuff (volume_util*.go) somewhere downstream. I added this code to [commit `ff29c85`](https://github.com/openshift/local-storage-operator/pull/431/commits/ff29c851192776f5c89c0f807f2f8a55cfebb810), so we have "unsupported" message at build time now.

diskmaker/controllers/lv/reconcile_test.go
mpatlasov: OK, agreed, let's not hide the errors: https://github.com/openshift/local-storage-operator/pull/431/commits/ff29c851192776f5c89c0f807f2f8a55cfebb810#diff-043dd3dc2a6dd72e8b4213a58c4ca5b3a9361dbc44f8313756306dced58caf8dR36-R38

mpatlasov:
Last push improved tests per Roman's suggestions.

chao007:
oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.15.0-0.test-2023-10-13-020623-ci-ln-l1brb62-latest   True        False         146m    Cluster version is 4.15.0-0.test-2023-10-13-020623-ci-ln-l1brb62-latest


```
sh-5.1# truncate -s 1G /tmp/img
sh-5.1# losetup /dev/loop0 /tmp/img
sh-5.1# mkfs.xfs /dev/loop0
meta-data=/dev/loop0             isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1    bigtime=1 inobtcount=1
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
Discarding blocks...Done.

oc get localvolume/local-disks -o yaml
apiVersion: local.storage.openshift.io/v1
kind: LocalVolume
metadata:
  creationTimestamp: "2023-10-13T04:32:14Z"
  finalizers:
  - storage.openshift.com/local-volume-protection
  generation: 4
  name: local-disks
  namespace: openshift-local-storage
  resourceVersion: "76173"
  uid: f31894cd-b3f8-4bbd-bb38-d9bae261752a
spec:
  logLevel: Normal
  managementState: Managed
  nodeSelector:
    nodeSelectorTerms:
    - matchExpressions:
      - key: kubernetes.io/hostname
        operator: In
        values:
        - ci-ln-l1brb62-72292-2d6mc-worker-a-sx8vs
  storageClassDevices:
  - devicePaths:
    - /dev/loop0
    - /dev/loop24
    forceWipeDevicesAndDestroyAllData: true
    fsType: ext4
    storageClassName: local-sc
    volumeMode: Filesystem
status:
  conditions:
  - lastTransitionTime: "2023-10-13T04:32:14Z"
    message: Ready
    status: "True"
    type: Available
  generations:
  - group: apps
    lastGeneration: 3
    name: diskmaker-manager
    namespace: openshift-local-storage
    resource: DaemonSet
  managementState: Managed
  observedGeneration: 4
  readyReplicas: 0

oc rsh fedora0
sh-5.2$ df /mnt/claim0 
Filesystem     1K-blocks  Used Available Use% Mounted on
/dev/loop0        996780    24    980372   1% /mnt/claim0
sh-5.2$ grep "/dev/loop0" /proc/mounts 
/dev/loop0 /mnt/claim0 ext4 rw,seclabel,relatime 0 0
```

PR #430, opened 2023-09-26
STOR-1402: Chore: update libraries in all operators

dobsonj:
https://issues.redhat.com/browse/STOR-1402

This bumps all libraries for OCP 4.15 / k8s 1.28.
controller-runtime removed some deprecated fields that LSO was still using,
which required some changes in main, diskmaker_manager, and unit tests.

cc @openshift/storage

PR #433, opened 2023-09-28
STOR-1462: CRD manifests should be generated automatically

dobsonj:
https://issues.redhat.com/browse/STOR-1462

I did quite a lot of cleanup work in this PR outside the scope of the original ticket. It's best to review the commits individually. I left them separate to make reviews easier, and I can split some of these changes out if needed.

The intent overall is to make it easier to contribute to LSO and remove artifacts that are not needed for building, developing, or testing LSO.

Some overall themes:
- Building bundles has long been error prone, requiring manual steps. Now you can just `make bundle REGISTRY=quay.io/username` and that's it. See HACKING.md for more details.
- Makefile was full of cruft, and there were duplicate manifests, unused manifests, and outdated examples. I removed anything that was not obviously useful.
- The remaining manifests in config/crd and config/rbac are generated now by `make update`. The generated CRD's are included in the bundle. The generated RBAC rules are incomplete, so the manifests in config/rbac are there for reference only right now. We should do some follow-up work on this in another PR.
- `make verify` ensures we don't forget to run `make update` -- I'll add a verify presubmit job after this PR merges. I fixed up all verify errors so it runs cleanly.
- I moved all go code to either pkg/ cmd/ or test/ and updated all package references. This is just to make the code organization easier to reason through and be consistent with other operators.

cc @openshift/storage


dobsonj:
`image-registry.openshift-image-registry.svc:5000/ci-op-z4j69dnp/pipeline: received unexpected HTTP status: 500 Internal Server Error`
/retest-required

mpatlasov:
Last commit ("replace hack/sync_bundle, fix make bundle, remove dead dockerfiles") would collide with Roman's https://github.com/openshift/local-storage-operator/pull/410. @RomanBednar , you re-wokred sync_bundle long time ago (in June?). Jonathan's bash script ("create-bundle") looks more readable than go version (for my eyes). Do you have any preferences? (maybe your go prog does something important that bash script misses?)

mpatlasov:
Side note for reviewers: the first commit ("STOR-1462: CRD manifests should be generated automatically") modifies four yaml-s in `config/crd/bases/`. These changes didn't come from human, these files were auto-generated by newer version of `controller-gen` and its results differ from the older version of `controller-gen`. (somehow it took long while for me to discover that)

dobsonj:
I just pushed several new commits and updated the PR description, but I think I'm done updating this now (aside from addressing review comments).

I'm partial to the `create-bundle` bash script because:
- It's a lightly modified version of what we have for [EFS](https://github.com/openshift/aws-efs-csi-driver-operator/blob/master/hack/create-bundle), [filestore](https://github.com/openshift/gcp-filestore-csi-driver-operator/blob/main/hack/create-bundle), and [secrets-store](https://github.com/openshift/secrets-store-csi-driver-operator/blob/main/hack/create-bundle). i.e. it's consistent with all our other OLM operators.
- It introduces very little code and no external dependencies aside from opm and docker / podman. The less code we have to maintain, the better.

mpatlasov:
`ci/prow/e2e-operator-extended` is perma-failing:

```
: Run multi-stage test e2e-operator-extended - e2e-operator-extended-openshift-extended-test container test expand_less | 11m17s
-- | --
{  e8x Sep 30 00:12:56.122 W persistentvolume/local-pv-eef239ed Error cleaning PV "local-pv-eef239ed": Unknown storage class name lvsc-j0g3se8x (2 times) Sep 30 00:13:09.406 W persistentvolume/local-pv-eef239ed Error cleaning PV "local-pv-eef239ed": Unknown storage class name lvsc-j0g3se8x (3 times) Sep 30 00:13:09.410 I persistentvolume/local-pv-f0173d8c Starting cleanup of Block PV "local-pv-f0173d8c", this may take a while Sep 30 00:13:17.105 W persistentvolume/local-pv-f0173d8c Error cleaning PV "local-pv-f0173d8c": Unknown storage class name lvsc-t7kmmne9 Sep 30 00:13:17.111 W persistentvolume/local-pv-f0173d8c Error cleaning PV "local-pv-f0173d8c": Unknown storage class name lvsc-t7kmmne9 (2 times) Sep 30 00:13:17.121 W persistentvolume/local-pv-f0173d8c Error cleaning PV "local-pv-f0173d8c": Unknown storage class name lvsc-t7kmmne9 (3 times) Sep 30 00:13:17.153 W persistentvolume/local-pv-f0173d8c Error cleaning PV "local-pv-f0173d8c": Unknown storage class name lvsc-t7kmmne9 (4 times) Sep 30 00:13:17.158 W persistentvolume/local-pv-f0173d8c Error cleaning PV "local-pv-f0173d8c": Unknown storage class name lvsc-t7kmmne9 (5 times) Sep 30 00:14:07.643 I persistentvolume/local-pv-eef239ed Starting cleanup of Block PV "local-pv-eef239ed", this may take a while Sep 30 00:14:17.671 I persistentvolume/local-pv-f0173d8c Starting cleanup of Block PV "local-pv-f0173d8c", this may take a while  Writing JUnit report to /logs/artifacts/junit/junit_e2e_20230930-001453.xml  8 pass, 0 skip (10m47s) + set +x try to handle result /logs/artifacts/junit/junit_e2e_20230930-001453.xml /logs/artifacts/junit/junit_e2e_2023-09-30-00-14-53.xml done to handle result success force success exit {"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"k8s.io/test-infra/prow/entrypoint/run.go:84","func":"k8s.io/test-infra/prow/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2023-09-30T00:14:53Z"} error: failed to execute wrapped command: exit status 1 }
```
(https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_local-storage-operator/433/pull-ci-openshift-local-storage-operator-master-e2e-operator-extended/1707897311264247808)

See also this comment: https://github.com/openshift/local-storage-operator/pull/435#issuecomment-1742347599

dobsonj:
latest push is just a rebase after https://github.com/openshift/local-storage-operator/pull/430 merged

RomanBednar:
> Last commit ("replace hack/sync_bundle, fix make bundle, remove dead dockerfiles") would collide with Roman's #410. @RomanBednar , you re-wokred sync_bundle long time ago (in June?). Jonathan's bash script ("create-bundle") looks more readable than go version (for my eyes). Do you have any preferences? (maybe your go prog does something important that bash script misses?)

@mpatlasov That is correct, thanks for pointing that out. Both scripts are essentially the same, the aim is to replace the error prone script that we had. There are a few differences:

- checking for `opm` binary - without it this script will build images (which takes a while) and can fail when it tries to run `opm` if not installed - this can be a bit annoying so I had a check for that
- checking container engine - this script is better that it considers both `docker` and `podman`
- this script is for creating bundle and index only, LSO images are built and pushed via makefile. I intended to have standalone script that does everything on one place - this is a matter of preference
- with this version you're going to need 5 different quay.io namespaces set to be public (`local-storage-operator`, `local-storage-diskmaker`, etc...). I assumed that most of us use just one for testing and use tag for naming instead of version - again, not too important
- `sed` is platform specific, so that's one advantage of not using bash, on macOS I get this when I run the script:

```
sed: -I or -i may not be used with stdin
make: *** [bundle] Error 1
```

Bash is way more readable and consistent with other repos, so we can stick with that (maybe with some small improvements).

dobsonj:
Thanks for the review and feedback @RomanBednar , I pushed another commit to add the `opm` check and to fix the `sed` command on BSD/MacOS. Could you try the latest version to make sure it works on your mac?

> this script is for creating bundle and index only, LSO images are built and pushed via makefile. I intended to have standalone script that does everything on one place - this is a matter of preference

I guess this is just preference, but I like the makefile targets just because it makes it easy to rebuild a single image when needed (instead of the whole bundle).

> with this version you're going to need 5 different quay.io namespaces set to be public (local-storage-operator, local-storage-diskmaker, etc...). I assumed that most of us use just one for testing and use tag for naming instead of version - again, not too important

This is a good point, it's a little cumbersome to setup at first. I do like having a version tag though, sometimes it's useful to build images for a specific bug fix and leave it, without getting overwritten by other dev changes.

A middle ground could be something like:
```
quay.io/user/lso:operator-version
quay.io/user/lso:diskmaker-version
quay.io/user/lso:mustgather-version
quay.io/user/lso:bundle-version
quay.io/user/lso:index-version
```
@RomanBednar @mpatlasov what do you guys think, would this change be useful?

dobsonj:
> > with this version you're going to need 5 different quay.io namespaces set to be public (local-storage-operator, local-storage-diskmaker, etc...). I assumed that most of us use just one for testing and use tag for naming instead of version - again, not too important
> 
> This is a good point, it's a little cumbersome to setup at first. I do like having a version tag though, sometimes it's useful to build images for a specific bug fix and leave it, without getting overwritten by other dev changes.
> 
> A middle ground could be something like:
> 
> ```
> quay.io/user/lso:operator-version
> quay.io/user/lso:diskmaker-version
> quay.io/user/lso:mustgather-version
> quay.io/user/lso:bundle-version
> quay.io/user/lso:index-version
> ```

I pushed [another commit](https://github.com/openshift/local-storage-operator/pull/433/commits/9f633da964c3fd66b861c8436e1e51d8e4fe8db9) to do this ^
But I also have no problem dropping this commit if anyone objects to it.

gnufied:
Running `make` results in following modified files:

```
        modified:   config/crd/bases/local.storage.openshift.io_localvolumediscoveries.yaml
        modified:   config/crd/bases/local.storage.openshift.io_localvolumediscoveryresults.yaml
        modified:   config/crd/bases/local.storage.openshift.io_localvolumes.yaml
        modified:   config/crd/bases/local.storage.openshift.io_localvolumesets.yaml
        modified:   config/rbac/diskmaker/role.yaml
        modified:   config/rbac/role.yaml
```

With diff being something on the line of:

```
--- a/config/crd/bases/local.storage.openshift.io_localvolumediscoveries.yaml
+++ b/config/crd/bases/local.storage.openshift.io_localvolumediscoveries.yaml
@@ -3,7 +3,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.12.1
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
   name: localvolumediscoveries.local.storage.openshift.io
 spec:
   group: local.storage.openshift.io
@@ -114,12 +115,10 @@ spec:
                             type: object
                           type: array
                       type: object
-                      x-kubernetes-map-type: atomic
                     type: array
                 required:
                 - nodeSelectorTerms
                 type: object
-                x-kubernetes-map-type: atomic
               tolerations:
                 description: If specified tolerations is the list of toleration that
                   is passed to the LocalVolumeDiscovery Daemon
@@ -200,3 +199,9 @@ spec:
     storage: true
     subresources:
       status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
diff --git a/config/crd/bases/local.storage.openshift.io_localvolumediscoveryresults.yaml b/config/crd/bases/local.storage.openshift.io_localvolumediscoveryresults.yaml
index 2f2bf3b1..0e00d4c4 100644
--- a/config/crd/bases/local.storage.openshift.io_localvolumediscoveryresults.yaml
+++ b/config/crd/bases/local.storage.openshift.io_localvolumediscoveryresults.yaml
@@ -3,7 +3,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.12.1
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
   name: localvolumediscoveryresults.local.storage.openshift.io
 spec:
   group: local.storage.openshift.io
@@ -121,3 +122,9 @@ spec:
     storage: true
     subresources:
       status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
diff --git a/config/crd/bases/local.storage.openshift.io_localvolumes.yaml b/config/crd/bases/local.storage.openshift.io_localvolumes.yaml
```

I am guessing this is because of version difference between `controller-gen` binaries?

dobsonj:
> I am guessing this is because of version difference between `controller-gen` binaries?

Yes that's right, if I change controller-gen back to v0.8.0 it completely reverts those CRD changes.

```
$ git diff
diff --git a/Makefile b/Makefile
index d66a660a..9e3ca2dd 100644
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ test: ## Run unit tests.

 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-       $(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.12.1)
+       $(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)

 # go-get-tool will 'go get' any package $2 and install it to $1.
 PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
$ rm bin/controller-gen
$ make update
```

I didn't change the controller-gen version in this PR, but there was no Makefile target that updated those CRD's before.

dobsonj:
> I am guessing this is because of version difference between `controller-gen` binaries?

Oh I see, you have v0.8.0 still in your workspace.
`rm bin/controller-gen` and try `make update` again.

dobsonj:
I restored the `config/samples` manifests in the `remove unused rbac manifests from config dir` commit so operator-sdk will not complain about missing examples:
```
$ operator-sdk generate bundle --input-dir config
Generating bundle
Generating bundle manifests
Bundle manifests generated successfully in bundle
Generating bundle metadata
INFO[0000] Creating bundle.Dockerfile                   
INFO[0000] Creating bundle/metadata/annotations.yaml    
INFO[0000] Bundle metadata generated successfully       

$ operator-sdk bundle validate ./bundle
INFO[0000] All validation tests have completed successfully 
```

HACKING.md
```diff
@@ -42,97 +42,60 @@ oc scale --replicas=0 deployment.apps/local-storage-operator -n openshift-local-
...
+After pushing all the images, it will print a subscription you can copy and paste to install the catalog source on your cluster:
 
 ```
-~> oc create -f examples/olm/catalog-create-subscribe.yaml
-```
-
-## Manual creation of bundle and index image.
-
-1. Since we will be going to test with our version of images, we need to modify CSV file to point to our version of image. This can be done by modifying following file:
+Copy following snipped to apply it to your cluster
```
RomanBednar, line 67:
```suggestion
Copy following snippet to apply it to your cluster
```

hack/create-bundle
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh
```
RomanBednar, line 1: nit: other bash scripts here have `.sh` suffix so it should be here as well for consistency

Makefile
```diff
@@ -159,106 +92,51 @@ rm -rf $$TMP_DIR ;\
...
 images: diskmaker-container operator-container must-gather
+.PHONY: images
 
-push: images push-images
-
-push-images:
-	docker push ${DISKMAKER_IMAGE}
-	docker push ${OPERATOR_IMAGE}
-	docker push ${MUST_GATHER_IMAGE}
+push: images
```
RomanBednar, line 109: This will trigger must-gather image build from `Dockerfile.mustgather` which refers to an image not available on arm:

```
docker pull registry.ci.openshift.org/ocp/4.15:cli
cli: Pulling from ocp/4.15
no matching manifest for linux/arm64/v8 in the manifest list entries
```

The image (`registry.ci.openshift.org/ocp/4.15:must-gather`) that was referenced in removed `Dockerfile.mustgather.rhel7` works fine but I don't know if they're same or not. Maybe the one tagged `must-gather` is a replacement for the `cli` one?

HACKING.md
dobsonj: fixed (here and in hack/create-bundle.sh)

hack/create-bundle
dobsonj: done

Makefile
dobsonj: Since this is the only image having issues cross compiling, and we don't normally need to rebuild it, I dropped it as a requirement for `make bundle`. So the bundle will just use `quay.io/openshift/origin-local-storage-mustgather:latest` for the mustgather image (same as `hack/sync_bundle` did).

dobsonj: As far as `4.15:cli` vs `4.15:must-gather` goes, it should be using `4.15:cli` as the base image. The ART bot changed the base image to `cli` back in https://github.com/openshift/local-storage-operator/commit/03f0cf29c4d19ff01cda8ca5350dc137872ed3b8 -- Dockerfile.mustgather.rhel7 just never got updated because it's not used or maintained by ART.

mpatlasov:
@dobsonj , the first commit in this PR replaces 4 yaml-s we had in `config/manifests/stable/` with 4 yaml-s auto-generated by new `manifests` target. I attempted to compare what we had vs. what we generate now and it's really cumbersome because both indentation and order changed. However, I discovered one discrepancy:

Older `config/manifests/stable/local-volume-discoveries.crd.yaml` had:
```
          required:
              - spec
```
in `spec.versions[0].schema.openAPIV3Schema` while newer `config/crd/bases/local.storage.openshift.io_localvolumediscoveries.yaml` doesn't have it. Two questions about it:

1) How did you decide that that discrepancy is unimportant?
2) How did you compare those 4 yaml-s (by eyes or by some tooling)?

Makefile
```diff
@@ -48,63 +14,49 @@ TARGET_GOARCH=$(shell go env GOARCH)
...
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./api/..."
+.PHONY: generate
 
 fmt: ## Run go fmt against code.
 	go fmt ./...
+.PHONY: fmt
 
-# vet: ## Run go vet against code.
-# 	go vet ./...
+vet: ## Run go vet against code.
```
mpatlasov, line 64: This belongs to `16baf4a4` ("add go vet to verify target and fix go vet errors"), not to this commit. At this point nobody uses this target, and `go vet` reports errors anyway.

hack/verify-manifests.sh
```diff
@@ -6,6 +6,6 @@ trap "test -d $tmpdir && rm -rf $tmpdir" EXIT
 echo "Backup config directory to $tmpdir first"
 cp -r config $tmpdir/config
 
-make update
+make rbac manifests
```
mpatlasov, line 9: This makes filename misleading. It must be `hack/verify-rbac-and-manifests.sh` then.

Makefile
```diff
@@ -123,13 +130,13 @@ test: manifests generate diskmaker-rbac fmt ## Run tests.
 
 ##@ Deployment
 
-install: manifests diskmaker-rbac kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+install: manifests rbac kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
```
mpatlasov: Useless to fix what you remove altogether in the next commit.

Makefile
```diff
@@ -123,13 +130,13 @@ test: manifests generate diskmaker-rbac fmt ## Run tests.
...
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
-uninstall: manifests diskmaker-rbac kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
+uninstall: manifests rbac kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
```
mpatlasov: Useless to fix what you remove altogether in the next commit.

Makefile
```diff
@@ -123,13 +130,13 @@ test: manifests generate diskmaker-rbac fmt ## Run tests.
...
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
-deploy: manifests diskmaker-rbac kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+deploy: manifests rbac kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
mpatlasov: Useless to fix what you remove altogether in the next commit.

mpatlasov:
@dobsonj , the commit `b654e701f` ("remove unused kustomize crd patches") looks reasonable, but how did you find out that nobody uses that kustomize stuff?

mpatlasov:
@dobsonj , why are we interested in BSD at all? (`3bb170c`: create-bundle: add opm check and fix sed command on BSD)

HACKING.md
```diff
@@ -42,97 +42,60 @@ oc scale --replicas=0 deployment.apps/local-storage-operator -n openshift-local-
 All the images including operator, diskmaker, bundle and index images can be created in one shot using following command:
 
 ```
-~> ./hack/sync_bundle -o <operator_image> -d <diskmaker_image> -b <bundle_image> -i <index_image> bundle
+make bundle REGISTRY=quay.io/username
 ```
 
-This command also pushes the images to selected docker registry, so if you ran this command with following arguments:
+You may optionally set `VERSION` to be used as a tag, and `TOOL_BIN` to specify which tool to use (defaults to podman or docker).
```
mpatlasov: Imagine that the user has both podman and docker installed, but he/she wants to use docker as in your example below:
```
make bundle REGISTRY=quay.io/username VERSION=latest TOOL_BIN=`which docker`
```
Then docker will be used only in Makefile, but [`hack/create-bundle`](https://github.com/openshift/local-storage-operator/pull/433/commits/8aa1d0729e319e4bddef7d9ea68bcf9dd738f4a9#diff-d9ff92edcb92263c6a6d63ab25bcd1abb0ad4825f599d479d31636840b118eb8R15) would still use podman.

I think it's not nice. We have to pass env var `TOOL_BIN` to the script and use it (if it is set).

mpatlasov:
Minor nit: last commit (`34e906c`: address review comments, remove mustgather from make bundle) is better to squash with previous ones. (Future code readers will be puzzled why you introduced mustgather in `hack/create-bundle` to remove it very soon)

Makefile
dobsonj: I squashed these commits

Makefile
dobsonj: True, but I would still like to keep this change in this commit, so at least it refers to a valid Makefile target (even though it's removed by the next commit).

Makefile
dobsonj: True, but I would still like to keep this change in this commit, so at least it refers to a valid Makefile target (even though it's removed by the next commit).

Makefile
dobsonj: True, but I would still like to keep this change in this commit, so at least it refers to a valid Makefile target (even though it's removed by the next commit).

hack/verify-manifests.sh
dobsonj: I split this into `verify-rbac.sh` and `verify-manifests.sh` instead, so they each have a clear function, and we could extend one without affecting the other in the future.

dobsonj:
> @dobsonj , the first commit in this PR replaces 4 yaml-s we had in `config/manifests/stable/` with 4 yaml-s auto-generated by new `manifests` target. I attempted to compare what we had vs. what we generate now and it's really cumbersome because both indentation and order changed. However, I discovered one discrepancy:
> 
> Older `config/manifests/stable/local-volume-discoveries.crd.yaml` had:
> 
> ```
>           required:
>               - spec
> ```
> 
> in `spec.versions[0].schema.openAPIV3Schema` while newer `config/crd/bases/local.storage.openshift.io_localvolumediscoveries.yaml` doesn't have it. Two questions about it:
> 
> 1. How did you decide that that discrepancy is unimportant?
> 2. How did you compare those 4 yaml-s (by eyes or by some tooling)?

I compared the yamls only with my eyes and otherwise relied on testing. There are a lot of formatting differences, but I did not identify any functional differences.


> @dobsonj , the commit `b654e701f` ("remove unused kustomize crd patches") looks reasonable, but how did you find out that nobody uses that kustomize stuff?
> 
> The same question is about next commit -- `5b88fc0` ("remove dead manifests under config dir")

The kustomize patches were mostly boilerplate manifests generated by operator-sdk, and they were never used to generate the manifests that get deployed as part of the bundle.

I grepped for each filename to find any existing references in the repo, and also looked for these objects on a cluster with LSO deployed. If they are not referenced by any code, makefile targets, or docs and they are not deployed on the cluster and they do not have a clearly defined purpose, then there isn't much reason to keep them.


> @dobsonj , why are we interested in BSD at all? (`3bb170c`: create-bundle: add opm check and fix sed command on BSD)

We're not interested in BSD per se, but this commit was to fix the issue Roman pointed out when running create-bundle.sh on MacOS (which uses the BSD version of sed). I squashed that commit though in my latest push, so it's part of the create-bundle commit now.


> Minor nit: last commit (`34e906c`: address review comments, remove mustgather from make bundle) is better to squash with previous ones. (Future code readers will be puzzled why you introduced mustgather in `hack/create-bundle` to remove it very soon)

Fixed, I squashed related commits.

@mpatlasov @RomanBednar thanks very much for the reviews, I think I addressed outstanding comments.

HACKING.md
dobsonj: Yes good catch! I addressed this in the last commit.

dobsonj:
no end-user impact or doc changes
/label px-approved
/label docs-approved

Phaow:
Go through the commits, looks good. Also CI results looks good.
/label qe-approved

PR #443, opened 2023-10-23
OCPBUGS-22250: Use mkfs.xfs in quick_reset.sh

mpatlasov:
mkfs has a limit of 16TB, let's use mkfs.xfs to avoid failure on large volumes.

### DIFF
--- a/hack/scripts/quick_reset.sh
+++ b/hack/scripts/quick_reset.sh
@@ -32,7 +32,8 @@ fi
 validateBlockDevice
 
 echo "Calling mkfs"
-ionice -c 3 mkfs -F $LOCAL_PV_BLKDEVICE
+# mkfs.xfs is preferred because default mkfs has 16TB limit (https://access.redhat.com/solutions/66689)
+ionice -c 3 mkfs.xfs -f $LOCAL_PV_BLKDEVICE
 
 echo "Calling wipefs"
 ionice -c 3 wipefs -a $LOCAL_PV_BLKDEVICE


mpatlasov:
Last push added `blkdiscard` call to quick_reset.sh

hack/scripts/quick_reset.sh
```diff
@@ -31,8 +31,12 @@ fi
 # Validate that we got a valid block device to cleanup
 validateBlockDevice
 
+echo "Calling blkdiscard (failure is permitted -- some devices do not support discard)"
+ionice -c 3 blkdiscard -f $LOCAL_PV_BLKDEVICE || true
+
 echo "Calling mkfs"
-ionice -c 3 mkfs -F $LOCAL_PV_BLKDEVICE
+# mkfs.xfs is preferred because default mkfs has 16TB limit (https://access.redhat.com/solutions/66689)
+ionice -c 3 mkfs.xfs -f $LOCAL_PV_BLKDEVICE
```
jsafrane, line 39: * `blkdiscard` above marks all blocks as unused, so lower storage layers may re-use the blocks. This is just to be nice to the underlying device and may not discard anything on devices that don't support it.
* `wipefs` below destroys headers of *known* filesystems, partition tables and similar things, so the device appears as it does not have any.

So why do we need `mkfs.xfs`, that destroys the first 512 bytes (overwrites them with xfs superblock) + puts xfs metadata on the disk and discards all other blocks (if supported by the device)? One reason could be to destroy header of unknown data on the device, when it is used as a raw block and e.g. a database is stored there without any filesystem. We should either add a comment about this use case or replace it with something better (`dd` the first block?)

Alveel: For discard support detection, would [this](https://unix.stackexchange.com/a/668323) be helpful? Seems very simple to implement.

I agree that doing a `mkfs` at all seems a bit weird. A simple `dd` should do the trick and not require a `wipefs` call anymore. I saw an implementation for this [upstream](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/deployment/docker/scripts/dd_zero.sh), though I'm sure you've also already seen it.

mpatlasov: I've just double-checked the behavior of mkfs.xfs and observed that it always discard all blocks from the device (if discard is supported).

mpatlasov: @Alveel, full `dd` is too heavy (would take too long), and using mkfs gives us some trade off clearing at least some amount of data. This PR simply replaces mkfs.ext2 with mkfs.xfs. I agree that mkfs seems a bit weird, but it corresponds to upstream and it's enough to fix OCPBUGS-22250.

mpatlasov:
Last push removed explicit `blkdiscard` call because `mkfs.xfs` always attempts to discard everything anyway.

hack/scripts/quick_reset.sh
Alveel: Of course, I think you'd only want to `dd` the first few blocks. The upstream script allows for that :)

Regardless, happy to see this bug fixed this quickly.

PR #466, opened 2024-01-19
STOR-1574: Bump library-go

jsafrane:
Update openshift/library-go and openshift/api to today's master. Update k8s deps to v0.29.0. Manually bumped controller-runtime.

@openshift/storage


jsafrane:
/label px-approved
/label docs-approved
No changes from user perspective
/label qe-approved
CI should be enough for library bumps

PR #475, opened 2024-03-26
OCPBUGS-26028: local-storage-operator CSV should set spec.replaces

dobsonj:
/cc @openshift/storage


dobsonj:
/hold
operator-e2e is failing for efs, filestore, secrets-store operator PR's, and the approach is the same.

dobsonj:
This PR causes a bundle error, and it's no longer needed since ART is working on cleaning up the 4.15 images that cause this test to report a failure.

PR #494, opened 2024-07-26
STOR-773: LSO and static-provisioner library reconcile

dobsonj:
https://issues.redhat.com/browse/STOR-773

Goal:
- Restructure diskmaker so we can drop [this carry patch](https://github.com/openshift/sig-storage-local-static-provisioner/commit/b5e1aa52a4e3269c1a709c6018d00877d18991eb) and still avoid reintroducing [BZ 2032924](https://bugzilla.redhat.com/show_bug.cgi?id=2032924).

Approach:
- Serialize deletion and creation of PV's [like local-static-provisioner does](https://github.com/openshift/sig-storage-local-static-provisioner/blob/e8e0708020d5c2a1da8704be2f12afa8d3c87b0c/pkg/controller/controller.go#L118-L120) to avoid competing delete / create requests. The LV controller and LVS controller will handle their respective PV deletion, and the deleter controller can be completely removed.
- Do not remove the cleanup job in CreateLocalPV(), this causes the PV [not to be deleted](https://github.com/openshift/sig-storage-local-static-provisioner/blob/b5e1aa52a4e3269c1a709c6018d00877d18991eb/pkg/deleter/deleter.go#L170)! Instead we should trigger a fast requeue to call DeletePVs() again while the volume is released, which will check the cleanup job status and delete the PV when cleanup is complete.
- Remove the Sleep() that was in handlePVChange(), this doesn't do what it looks like. `isDelete` is only true _after_ the PV has been deleted. At that point, the cleanup job is already removed.
- Refactor the code and simplify where possible.

Testing:
- [x] Run the reproducer from [BZ 2032924](https://bugzilla.redhat.com/show_bug.cgi?id=2032924) and monitor diskmaker logs to ensure this use case still works.
- [x] Unit tests / move pkg/diskmaker/controllers/deleter/reconcile_test.go
- [x] Run reproducer script for LVSet objects
- [x] Run concurrent create/delete requests for 4 PV's under a single LV
- [x] Run concurrent create/delete requests for 4 PV's under 4 different LV's
- [x] Run reproducer scripts for LV and LVSet at the same time on the same node

/cc @openshift/storage


bertinatto (commented):
I reviewed each of the commits individually and addressed my questions with Jonathan  (that's why there are no questions in the diff).

Given my limited experience with LSO, the changes here seem to be a significant improvement over the existing state.

Thanks for working on this, @dobsonj!

/lgtm

PR #513, opened 2025-01-17
OCPBUGS-48322: Fast requeue when LV / LVS deletion is blocked

RomanBednar:
If users delete LV before deleting PVC LSO will put a finalizer on LV and start a cleanup job for the volume but won't reconcile again automatically to finish the cleanup and remove the finalizer and PV. Because of this a bulk deletion of resources with `oc delete` can hang forever, unless something else triggers reconcile.

This PR is basically a partial cherry-pick of https://github.com/openshift/local-storage-operator/commit/bd326bf1b3eba6af0632546b2fa5983671e4c68a to add a fast re-queue specifically for this case.

With this patch we can delete the resources in any order without getting into what appears to be a "stuck" state:

```
$ oc create -f /tmp/resources.yaml
namespace/mariadb-db created
localvolume.local.storage.openshift.io/local-disks created
persistentvolumeclaim/mariadb-persistent-storage created
configmap/mariadb-custom-config created
deployment.apps/mariadb-deployment created
service/mariadb-deployment created

$ oc -n mariadb-db get pod
NAME                                  READY   STATUS    RESTARTS   AGE
mariadb-deployment-5c6dc5b888-brc66   1/1     Running   0          23s

$ oc delete -f /tmp/resources.yaml
namespace "mariadb-db" deleted
localvolume.local.storage.openshift.io "local-disks" deleted
persistentvolumeclaim "mariadb-persistent-storage" deleted
configmap "mariadb-custom-config" deleted
deployment.apps "mariadb-deployment" deleted
service "mariadb-deployment" deleted
```

cc @openshift/storage

### DIFF
--- a/pkg/diskmaker/controllers/lv/reconcile.go
+++ b/pkg/diskmaker/controllers/lv/reconcile.go
@@ -324,7 +324,7 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 
 	// don't provision for deleted lvs
 	if !lv.DeletionTimestamp.IsZero() {
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: true, RequeueAfter: fastRequeueTime}, nil
 	}
 
 	klog.InfoS("Looking for valid block devices", "namespace", request.Namespace, "name", request.Name)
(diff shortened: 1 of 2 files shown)


dobsonj:
/lgtm
/label backport-risk-assessed

IMO, this is a very low-risk change. The problem in the bug is that cleanup completes, but because the LocalVolume has a deletionTimestamp, reconcile returns with no requeue. So unless there is some external event that triggers reconcile again, the loop never runs again to discover the completed cleanup job. This means the PV is not deleted, and the LocalVolume finalizer is not removed because there is still a PV in the Released state.

Adding requeue to this return statement causes reconcile to rerun again after 5 seconds, where it discovers the completed cleanup job and deletes the PV.

Phaow:
/label cherry-pick-approved
The CI looks good will do some test after it merged.

PR #523, opened 2025-03-14
STOR-1877: Run operators without root user control

dobsonj:
https://issues.redhat.com/browse/STOR-1877
/cc @openshift/storage


duanwei33:
CI results look good, which should be sufficient for merging from QE side.
/label qe-approved

PR #532, opened 2025-07-10
[release-4.18] OCPBUGS-59180: fix no warning for drift between ocp and lso versions

rhrmo:


RomanBednar:
@dobsonj This seems like an additional thing to remember when bumping OLM metadata, maybe we should consider a script for that or at least document it somewhere? 

A CI job to check this might be also easy to do, we can compare version in branch name with versions in metadata, `maxOpenShiftVersion` is always current release + 1 it seems...not sure it's worth the effort though.

PR #536, opened 2025-07-23
NO-JIRA: Add hack/update-metadata.sh script to bump OLM version

dobsonj:
Follow-up on the discussion in https://github.com/openshift/local-storage-operator/pull/532 to automate OLM version bumps.

```
$ make metadata
./hack/update-metadata.sh
Updating package manifest to 4.20.0
Updating OLM metadata to 4.20.0
```

/cc @openshift/storage @RomanBednar @rhrmo


dobsonj:
/hold
the [verify](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_local-storage-operator/536/pull-ci-openshift-local-storage-operator-main-verify/1948147821836242944/artifacts/test/build-log.txt) job is failing silently

Makefile
```diff
@@ -69,9 +76,15 @@ test: ## Run unit tests.
 	go test ./pkg/... -coverprofile cover.out
 .PHONY: test
 
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+YQ = $(BIN_PATH)/yq
```
RomanBednar: This is platform specific, can we reuse this or do something similar? https://github.com/openshift/build-machinery-go/blob/1b6d00b8c37cd8117556b99c800b453639b7f253/make/targets/openshift/yq.mk#L13-L22

🍎 😿 

There's some ancient yq version so it might need some tweaks. Or just get GOHOSTOS, GOHOSTARCH from go env to build the url here.

hack/update-metadata.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Usage:
```
RomanBednar, line 7: This was a bit unclear to me at first. The script can run in two modes, either no argument is provided and in that case it takes ocp version from `.channels[0].currentCSV` found in `local-storage-operator.package.yaml` and updates just the metadata accordingly, or it takes an argument with desired ocp version and updates both metadata **and** the `local-storage-operator.package.yaml` file.

This makes the semantics of the argument slightly unclear but I don't think it's worth investing more time in utility scripts like this one, however some nicer description would help our future selves.

README.md
```diff
@@ -6,3 +6,9 @@ Instructions to deploy on OCP >= 4.2 using OLM can be found [here](docs/deploy-w
 
 ## Using the must-gather image with the local storage operator
 Instructions for using the local storage's must-gather image can be found [here](docs/must-gather.md)
+
+## Bumping OCP version in OLM metadata
```
RomanBednar: This does not update just the metadata, but also CSV version in `local-storage-operator.package.yaml`. See my comment above for `update-metadata.sh` script.

Makefile
dobsonj: I knew something would break on Mac, thanks for testing this :joy: 
I updated this to reuse the ensure-yq target from build-machinery-go.

hack/update-metadata.sh
dobsonj: Yeah, I wanted the same script to be usable for `make update`, `make verify`, and also to manually set a specific version at the beginning of each release. I added some more details in the usage description here.

README.md
dobsonj: Ack, I updated the description to clarify what it changes.

Makefile
RomanBednar: Yeah, as usual :D now it's looking fine, thanks!

```
❯ make metadata
Installing yq into '/Users/MAC/git/local-storage-operator/bin/yq'
mkdir -p '/Users/MAC/git/local-storage-operator/bin/'
curl -s -f -L https://github.com/mikefarah/yq/releases/download/v4.47.1/yq_darwin_arm64 -o '/Users/MAC/git/local-storage-operator/bin/yq'
chmod +x '/Users/MAC/git/local-storage-operator/bin/yq';
./hack/update-metadata.sh
Updating package manifest to 4.20.0
Updating OLM metadata to 4.20.0
❯ echo $?
0
```

dobsonj:
IMO, the [verify](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_local-storage-operator/536/pull-ci-openshift-local-storage-operator-main-verify/1951061636953411584/artifacts/test/build-log.txt) job should be enough for this PR. There are no user-facing changes, just a script to simplify release bumps for maintainers.
/label qe-approved

PR #542, opened 2025-08-13
OCPBUGS-60033: Prefer nvme-eui symlinks over nvme-

jsafrane:
On vSphere, symlink `/dev/disk/by-id/nvme-VMware_Virtual_NVMe_Disk_VMware_NVME_0000` is not stable. It get reordered when a disk is added/removed to/from a VM + the VM is rebooted.

Prefer using `/dev/disk/by-id/nvme-eui.XYZ` if it's available.


jsafrane:
cc @openshift/storage

PR #594, opened 2026-03-19
wip: Add e2e test that checks PreferredLinkTarget is updated

jsafrane:
**See the last commit only, the rest is https://github.com/openshift/local-storage-operator/pull/592**

Update LocalVolume + LocalVolumeSets with a check for PreferredLinkTarget update.
    
The check picks one PV / LVDL, on the corresponding node it adds a new symlink to the corresponding device that's more preferred than the current one, and expects that the LVDL gets updated. Then it removes the symlink + checks the old PreferredLinkTarget is restored.
    
The test expects the test disks are NVMe. It creates a symlink `scsi-1`, which is more preferable by LSO. It can be extended in the future to add `scsi-3`, `wwn-` etc to check which one is actually the preferred one.
    
I also added some common functionality to add / remove udev-like symlinks on nodes easily and updated `newNodeJob` to get hostname instead of the whole Node object - it saves one API call to get a Node from hostname label from PV `nodeAffinity`.

WIP:
* remove 30 minute timeout that's there just for debugging
* perhaps refactor the check to a single function. Now it's copied to LV and LVSet unit tests.
* fix LocalVolumes - the test fails, because LSO does not update PreferredLinkTarget for LVs that are not mounted.


jsafrane:
/close
it has been merged as part of other LSO LVDL PRs.

PR #599, opened 2026-03-31
OCPBUGS-52262: Allow filtering out devices by the device name

tsmetana:
This should fix the problem with LSO picking up (and corrupting) RBD devices on Virt clusters. The patch adds a new `deviceExclusionSpec` type that allows for filtering out devices based on device name pattern (e.g. `rbd*`, which should fix the referenced bug).


pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -346,6 +346,20 @@ DeviceLoop:
...
 			}
 		}
+
+		for name, excluder := range exclusionMap {
+			valid, err := excluder(blockDevice, lvset.Spec.DeviceExclusionSpec)
+			if err != nil {
+				klog.ErrorS(err, "exclusion error", "device",
+					blockDevice.Name, "filter", name)
+				valid = false
+				continue DeviceLoop
```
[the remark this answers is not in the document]
tsmetana: This follows the same pattern as for the inclusion filter / map in the same loop. If we would like to change this behaviour, it should be done for all the relevant cases, which is not really something in the scope of the intended change.

api/v1alpha1/localvolumeset_types.go
```diff
@@ -77,6 +77,17 @@ type DeviceInclusionSpec struct {
...
 
+// DeviceExclusionSpec holds the exclusion filter spec
+type DeviceExclusionSpec struct {
+	// DeviceNameFilter is a list of glob patterns. Devices whose kernel name (KName) matches
+	// any of these patterns are excluded from automatic provisioning.
+	// Glob patterns follow filepath.Match syntax: '?' matches any single non-separator character,
+	// '*' matches any sequence of non-separator characters.
+	// Example: ["rbd*"] excludes all rbd0, rbd1, etc.
+	// +optional
+	DeviceNameFilter []string `json:"deviceNameFilter,omitempty"`
```
gnufied, line 89: Add maxitems validation please.

tsmetana: OK. Hopefully this is fine.

gnufied:
I tested this manually in CRC environment and it appears to work.

PR #600, opened 2026-04-05
OCPBUGS-81715: Fix monitoring of udev using nsenter

gnufied:
https://redhat.atlassian.net/browse/OCPBUGS-81715

## Summary by CodeRabbit

* **Refactor**
  * Improved disk device monitoring with safer process management and better shutdown coordination
  * Emits clearer runtime warnings when required system capabilities aren’t available and skips monitoring gracefully
  * Streams diagnostic output from the device monitor into warnings for faster troubleshooting
  * More reliable and responsive device detection in containerized and isolated environments when supported
  * Better resource handling and graceful error recovery during device discovery and monitoring

### DIFF
--- a/pkg/diskmaker/discovery/monitor.go
+++ b/pkg/diskmaker/discovery/monitor.go
@@ -57,11 +57,20 @@ func udevBlockMonitor(c chan string, period time.Duration) {
 func rawUdevBlockMonitor(c chan string, matches, exclusions []string) {
 	defer close(c)
 
-	// stdbuf -oL performs line bufferred output
-	cmd := exec.Command("stdbuf", "-oL", "udevadm", "monitor", "-u", "-k", "-s", "block")
+	cmd := newUdevMonitorCommand()
+	if cmd == nil {
+		klog.Warning("udevadm based monitoring requires nsenter")
+		return
+	}
+
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		klog.Warningf("Cannot open udevadm stdout: %v", err)
+		return
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		klog.Warningf("Cannot open udevadm stderr: %v", err)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/diskmaker/discovery/monitor.go
```diff
@@ -57,24 +58,42 @@ func udevBlockMonitor(c chan string, period time.Duration) {
...
+	if err != nil {
+		klog.Warningf("Cannot open udevadm stderr: %v", err)
+		return
+	}
 
 	err = cmd.Start()
 	if err != nil {
 		klog.Warningf("Cannot start udevadm monitoring: %v", err)
 		return
 	}
```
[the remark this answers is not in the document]
gnufied: No the failure mode is fine. this container already runs with privileged permissions.

Dockerfile.diskmaker.rhel7
```diff
@@ -9,7 +9,7 @@ COPY --from=builder /go/src/github.com/openshift/local-storage-operator/_output/
 COPY --from=builder /go/src/github.com/openshift/local-storage-operator/hack/scripts /scripts
 COPY config/manifests /manifests
 
-RUN yum install -y e2fsprogs xfsprogs && yum clean all && rm -rf /var/cache/yum
+RUN yum install -y e2fsprogs xfsprogs systemd-udev && yum clean all && rm -rf /var/cache/yum
```
jsafrane: If you're using nsenter, then `udevadm` does not need to be installed inside the container.

pkg/diskmaker/discovery/monitor.go
```diff
@@ -98,6 +117,20 @@ func rawUdevBlockMonitor(c chan string, matches, exclusions []string) {
 	klog.Info("udevadm monitor finished")
 }
 
+func newUdevMonitorCommand() *exec.Cmd {
+	// In the discovery pod, udevadm can reject monitor requests when running in
+	// a container mount namespace. Enter host mount namespace when possible.
+	if _, statErr := os.Stat("/proc/1/ns/mnt"); statErr == nil {
+		if _, pathErr := exec.LookPath("nsenter"); pathErr == nil {
+			return exec.Command("nsenter", "--mount=/proc/1/ns/mnt", "--", "udevadm", "monitor", "-u", "-k", "-s", "block")
+		}
```
jsafrane: It should either discover very precisely if it's in a container or not or have a cmdline option to use `stdbuf` below (having `nsenter` as the default). The `/proc/1/ns/mnt`  check above will pass on any recent Linux, even on the host.

gnufied: in the CRC cluster at least running without nsenter doesn't work:

```
udevadm monitor stderr: Running in chroot, ignoring request
W0405 18:15:35.460151 2162284 discovery.go:103] disabling udev monitoring
```

You are right, we probably don't need udevadm in the container then.

jsafrane: Second thoughts: what's wrong with calling `nsenter` even on the host and for local development / debugging? It's not necessary, but it should work. Should we remove the nsenter check + stdbuf fallback below?

gnufied: I removed the check. PTAL

PR #607, opened 2026-04-14
OCPBUGS-82545: Fire metric for missing symlink on LocalVolume.

dfajmon:
I have added a user-specified `SpecifiedLinkTarget` field to our api so we know when to fire the metric when the path is invalid due to changes such as an OS reinstall.

This was tested on GCP cluster with 2 scenarios.

1. Create localvolume -> Change symlink to invalid one -> see metric firing -> update back to symlink -> metric back to 0 occurences
2. Create localvolume -> change symlink to invalid one -> metric shows problems ->remove localvolume and pv -> metric back to 0 occurences

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status reporting for device link resolution on disk-identification lookup failures with a best-effort status update to keep device link state accurate.
* **New Features**
  * Added a public block-device resolution utility that resolves by-id symlinks and returns device details including the by-id path.
* **Tests**
  * Added unit tests covering successful and failing device-resolution and status-update scenarios.

### DIFF
--- a/assets/templates/localmetrics/prometheus-rule.yaml
+++ b/assets/templates/localmetrics/prometheus-rule.yaml
@@ -70,3 +70,19 @@ spec:
             Local Storage operator has observed that pv {{ $labels.persistent_volume }} is using a symlink
             that is not optimum for stability between reboots and OS upgrades. Consider updating
             LocalVolumeDeviceLink policy to preferredSymlink to allow LSO to manage the symlink.
+    - name: lso_lv_missing_device_path
+      rules:
+      - alert: LSOLocalVolumeMissingDevicePath
+        expr: min_over_time(lso_lv_missing_device_path_count[5m]) > 0
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: "LocalVolume references a device path that does not exist on the node"
+          description: |
+            Local Storage Operator cannot find {{ $value }} device path(s) listed in a LocalVolume
+            for storage class {{ $labels.storageClass }} on node {{ $labels.nodeName }}.
+            This typically happens when a /dev/disk/by-id/ symlink is removed during an OS upgrade
+            (e.g. RHEL 9 to RHEL 10). The existing PersistentVolumes may still be usable, but LSO
+            cannot monitor or update their symlinks until the LocalVolume devicePaths are corrected.
+            Update the LocalVolume spec to reference a device path that currently exists on the node.
--- a/pkg/diskmaker/controllers/lv/reconcile.go
+++ b/pkg/diskmaker/controllers/lv/reconcile.go
@@ -371,6 +371,8 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 		r.processRejectedDevicesForDeviceLinks(ctx, ignoredDevices, diskConfig)
 	}
 
+	r.updateMissingDevicePathMetrics(diskConfig)
+
 	mountPointMap, err := common.GenerateMountMap(r.runtimeConfig)
 	if err != nil {
 		klog.ErrorS(err, "failed to generate mountPointMap")
@@ -422,6 +424,18 @@ func (r *LocalVolumeReconciler) processValidDevices(ctx context.Context, validDe
 
 		// update metrics for orphaned symlink devices
 		localmetrics.SetLVOrphanedSymlinksMetric(nodeName, storageClass, len(orphanSymlinkDevices))
+	}
+}
+
+func (r *LocalVolumeReconciler) updateMissingDevicePathMetrics(diskConfig *DiskConfig) {
+	for storageClass, disks := range diskConfig.Disks {
+		missing := 0
+		for _, devicePath := range disks.DevicePaths {
+			if _, err := r.fsInterface.evalSymlink(devicePath); err != nil {
+				missing++
+			}
+		}
+		localmetrics.SetLVMissingDevicePathMetric(nodeName, storageClass, missing)
 	}
 }
 
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -43,6 +43,11 @@ var (
 		Help: "Total symlinks that became orphan after updating the devicePaths in LocalVolume CR",
 	}, []string{"nodeName", "storageClass"})
 
+	metricLocalVolumeMissingDevicePaths = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "lso_lv_missing_device_path_count",
+		Help: "Total device paths in LocalVolume spec that cannot be resolved on the node (e.g. symlink removed after OS upgrade)",
+	}, []string{"nodeName", "storageClass"})
+
 	LVDMetricsList = []prometheus.Collector{
 		metricDiscoveredDevicesByLocalVolumeDiscovery,
 	}
@@ -54,6 +59,7 @@ var (
 		metricLocalVolumeSetDeletionTimestamp,
 		metricLocalVolumeProvisionedPVs,
 		metricLocalVolumeOrphanedSymlinks,
+		metricLocalVolumeMissingDevicePaths,
 	}
 )
 
@@ -102,4 +108,16 @@ func SetLVOrphanedSymlinksMetric(nodeName, storageClassName string, count int) {
 	metricLocalVolumeOrphanedSymlinks.
 		With(prometheus.Labels{"nodeName": nodeName, "storageClass": storageClassName}).
 		Set(float64(count))
+}
+
+func SetLVMissingDevicePathMetric(nodeName, storageClassName string, count int) {
+	metricLocalVolumeMissingDevicePaths.
+		With(prometheus.Labels{"nodeName": nodeName, "storageClass": storageClassName}).
+		Set(float64(count))
+}
+
+func LVMissingDevicePathGauge(nodeName, storageClassName string) prometheus.Gauge {
+	return metricLocalVolumeMissingDevicePaths.With(prometheus.Labels{
+		"nodeName": nodeName, "storageClass": storageClassName,
+	})
 }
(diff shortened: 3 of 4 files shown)


assets/templates/localmetrics/prometheus-rule.yaml
```diff
@@ -70,3 +70,22 @@ spec:
             Local Storage operator has observed that pv {{ $labels.persistent_volume }} is using a symlink
             that is not optimum for stability between reboots and OS upgrades. Consider updating
             LocalVolumeDeviceLink policy to preferredSymlink to allow LSO to manage the symlink.
+    - name: lso_spec_device_path_missing
+      rules:
```
gnufied, line 74: This change should not be necessary right?  There is already an alert for mismatches

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -480,6 +483,36 @@ func (r *LocalVolumeReconciler) resolveValidDeviceLocation(devicePath string, fo
...
+	var blockDevice internal.BlockDevice
+	for _, target := range lvdl.Status.ValidLinkTargets {
+		bd, err := internal.GetBlockDeviceFn(target)
+		if err == nil {
+			blockDevice = bd
+			break
+		}
+	}
+	if blockDevice == (internal.BlockDevice{}) {
+		klog.Warningf("device %q: found in LVDL cache but all valid link targets are broken or missing; PV remains accessible but LVDL status cannot be refreshed", devicePath)
```
gnufied, line 507: So, pv can't remain accessible at this point IMO. So the warning should say something like:

"Found matching LVDL for devicePath %s , but devicePath points to no valid device".

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -480,6 +483,36 @@ func (r *LocalVolumeReconciler) resolveValidDeviceLocation(devicePath string, fo
...
+
+	// Try each ValidLinkTarget to find a working block device. This is the
+	// failure path so the extra lsblk calls are acceptable.
+	var blockDevice internal.BlockDevice
+	for _, target := range lvdl.Status.ValidLinkTargets {
+		bd, err := internal.GetBlockDeviceFn(target)
+		if err == nil {
+			blockDevice = bd
+			break
+		}
```
gnufied, line 504: so, I noticed that you moved this one here from my original recommendation of keeping this in `pv_link_cache` file? I.e - `GetLVDL` returns both the block device and LVDL.  Was there any reason? I think both work fine, but I preferred keeping this in pv_link_cache file for reusing this code for other places where we don't update LVDL because of errors.

dfajmon: Realised just later that it should be there to be reusable.

pkg/common/pv_link_cache.go
```diff
@@ -200,6 +234,18 @@ func (l *LocalVolumeDeviceLinkCache) watchForEvents(informer cache.Informer) (k8
 	})
 }
 
+func (l *LocalVolumeDeviceLinkCache) GetDirectLVDLMatch(symlink string) (CurrentBlockDeviceInfo, bool) {
```
jsafrane: Why do we need this function to make a copy of the cache? It's called only to call `GetLVDLAndBlockDevice` on the result. 

`LocalVolumeDeviceLinkCache.GetLVDLAndBlockDevice()`  would do the same job, without exposing the cache internals outside.

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -480,6 +483,25 @@ func (r *LocalVolumeReconciler) resolveValidDeviceLocation(devicePath string, fo
 	return deviceLocation, true, nil
 }
 
+func (r *LocalVolumeReconciler) updatePVLinkStatusOnLookupFailure(ctx context.Context, devicePath string) error {
```
jsafrane: This should be `updateDeviceLinkForMissingDevice` with a comment when it should be called and what it does.

jsafrane (commented):
I am not convinced this PR is good. Yes, a new `preferredLinkTarget` appears in LVDL.`status`, metric `lso_device_link_mismatch` is exposed for the bad device + user gets an alert. But our documentation and the whole API design imply that the diskmaker will fix the symlink when user sets LVDL `spec.policy: PreferredLinkTarget`. And that simply does not happen. because the bad devicePath does not reach the code that fixes the symlinks.

You _could_ add `RecreateSymlinkIfNeeded` call to `updatePVLinkStatusOnLookupFailure`, however, I am not sure it belongs there. We already have code that fixes the symlink and I don't think it should be called from multiple places.

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -458,6 +458,9 @@ func (r *LocalVolumeReconciler) resolveValidDeviceLocation(devicePath string, fo
 	if strings.HasPrefix(devicePath, diskByIDPrefix) {
 		matchedDeviceID, matchedDiskName, err := r.findDeviceByID(devicePath)
```
jsafrane, line 473: Maybe an alternative solution: here the LocalVolume reconciler knows that devicePath does not exist and stops processing that device. 

Instead, it should find LVDL that uses this `devicePath`, check all its `validLinkTargets` and use one that _exists_ instead of `devicePath` here to get `matchedDeviceID` and `matchedDiskName`. And continue with those further instead of returning error.,

That should update LVDL with a new `preferredTargetLink`, potentially fix the symlink or raise metric + alert. But please check very carefuly.

assets/templates/localmetrics/prometheus-rule.yaml
jsafrane: I think we need a new metric. `lso_device_link_mismatch` reports when `currentLinktarget != preferredLinkTarget`. It does not say anything about LocalVolume spec.storageClassDevices[*].devicePaths`. User should know when this `devicePath` disappears.

Example:

* LocalVolume wants: `devicePaths: ["/dev/disk/by-id/scsi-1-foo"]`.
* LocalVolumeDeviceLink has: `spec.policy: PreferredLinkTarget` and thus a better symlink is used: `status.currentLinkTarget: /dev/disk/by-id/wwn-stable` and `status.preferredLinkTarget: /dev/disk/by-id/wwn-stable`
* Now `scsi-1-foo` disappears during upgrade to RHEL 10.

The PV works, `lso_device_link_mismatch` does not report any mismatch, but LSO stops monitoring the device (that's what this PR +/- fixes now) and even if it still monitors the device, user has no indication that  there is something wrong with it. LSO won't be able to re-create the PV. We need  a separate metric for that.

gnufied: Yes,  we will need a new metric not an alert rule.  But it depends on what we are deciding to do when policy is `PreferredSymlink` and existing `disk-by-id` path from `LV` object is gone. 

I am leaning towards fixing the symlink if possible *always*. Currently we are fixing the symlinks when PV is in-use, but the moment PV is deleted even if `policy` is `PreferredSymlink`, we will stop re-creating the PV. So if we always fixed the symlink and prevented throwing an error when `disk-by-id` path from LV is missing then we won't need a separate metric, because in that case there is no error condition as such.

pkg/diskmaker/controllers/lv/reconcile.go
gnufied: If we are going to fix the symlink, do we need a metric+alert? 

TBH we have already fixed the symlink while PV was in-use, the problem is , we are stopping PV re-creation if `devicePath` in PV becomes invalid and PV is deleted.  

May be the solution is - we should fix symlink even if PV is deleted and we are re-creating it.  But I think fixing the symlink while PV is unavailable may require https://redhat.atlassian.net/browse/OCPBUGS-83305

assets/templates/localmetrics/prometheus-rule.yaml
jsafrane: We're diving into tiny details - I think it would be a bit fragile. For the user, it might be hard to figure out what devicePath in LocalVolume corresponds to which PV.

But I agree, as long as LSO re-creates a PV, we should be _mostly_ fine. If this PR fixes that, we can close the bug.

pkg/common/pv_link_cache.go
gnufied: We could do that, but `GetDirectLVDLMatch` as implemented if also scanned for devices will hold the lock for far longer than necessary.  Or we will need more granular locking in the `LocalVolumeDeviceLinkCache.GetLVDLAndBlockDevice()` so as lock is released after reading from the cache and before scanning `/dev/disk/by-id` etc.  `GetLVDLAndBlockDevice` will also call `lsblk`, so I am always wary of an external command just hanging. We generally do not want a function to hang with locks held. 

 Decoupling, the lvdl match (via `CurrentBlockDeviceInfo`) allows the lock to be released from `pv_link_cache` while caller can do its own processing on copy it received. I don't know but it feels cleaner to reason about from concurrency/locking perspective. No messy - mutex lock/unlock between functions.

jsafrane: You can do the same in a LocalVolumeDeviceLinkCache function. `LocalVolumeDeviceLinkCache.GetLVDLAndBlockDevice()` can make itself a copy while locked + then unlock. It's an orthogonal problem.

gnufied: I personally think that - by keeping the lock in `GetDirectLVDLMatch` and releasing the lock via `defer`, the function is basically atomic and re-entrant and very easy to reason about. 

By separating the device finding concerns (using lsblk), we are keeping critical section isolated from non-critical section in a very clean way. I am not sure if this is a orthogonal problem, it is how you structure your code for concurrency. And if we are going to copy the map anyways, what is even the point of keeping everything in one function? Just easy to trip over concurrency issues.

pkg/localmetrics/localmetrics.go
```diff
@@ -54,6 +59,7 @@ var (
 		metricLocalVolumeSetDeletionTimestamp,
 		metricLocalVolumeProvisionedPVs,
 		metricLocalVolumeOrphanedSymlinks,
+		metricLocalVolumeMissingDevicePaths,
 	}
```
gnufied, line 63: can we add some tests for this?

PR #603, opened 2026-04-10
OCPBUGS-83306: Further cleanup lso work

gnufied:
Fixes https://redhat.atlassian.net/browse/OCPBUGS-83306

## Summary by CodeRabbit

* **Tests**
  * Added sibling-fallback tests, new reusable fixtures, and centralized test utilities (scoped mocks, temp-dir helpers, deterministic fake command executors).

* **Refactor**
  * Modernized internal collection handling and reorganized event-watcher/cache startup so readiness is signaled after initial sync and population.

* **Bug Fixes**
  * Improved symlink handling and provisioning status updates; removed reliance on stale symlink arguments.

PR #608, opened 2026-04-14
OCPBUGS-83413: Add node name to lvdl

gnufied:
Fixes https://redhat.atlassian.net/browse/OCPBUGS-83413

includes changes from https://github.com/openshift/local-storage-operator/pull/603

## Summary by CodeRabbit

* **New Features**
  * LocalVolumeDeviceLink objects now require a nodeName field (1–253 chars) to associate links with a specific node.

* **Enhancements**
  * Device link discovery and indexing are scoped to the local node, reducing cross-node interference.
  * Manager startup now validates presence of the local node identity to ensure node-scoped behavior.

* **Tests**
  * Unit and e2e tests expanded for multi-node/shared-by-id scenarios; new e2e reproducer, cleanup helpers, and assertions added.

### DIFF
--- a/api/v1/localvolumedevicelink_types.go
+++ b/api/v1/localvolumedevicelink_types.go
@@ -52,12 +52,12 @@ type LocalVolumeDeviceLinkSpec struct {
 	// +kubebuilder:validation:MaxLength=253
 	PersistentVolumeName string `json:"persistentVolumeName"`
 
-	// nodeName is the name of the Openshift node on which this
+	// nodeName is the name of the OpenShift node on which this
 	// LocalVolumeDeviceLink object exists
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
-	NodeName string `json:"nodeName"`
+	NodeName string `json:"nodeName,omitempty"`
 
 	// policy expresses how to manage symlinks for the device.
 	// "None" means no policy has been chosen, and will generate an alert if
--- a/api/v1/localvolumedevicelink_types.go
+++ b/api/v1/localvolumedevicelink_types.go
@@ -52,6 +52,13 @@ type LocalVolumeDeviceLinkSpec struct {
 	// +kubebuilder:validation:MaxLength=253
 	PersistentVolumeName string `json:"persistentVolumeName"`
 
+	// nodeName is the name of the Openshift node on which this
+	// LocalVolumeDeviceLink object exists
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	NodeName string `json:"nodeName"`
+
 	// policy expresses how to manage symlinks for the device.
 	// "None" means no policy has been chosen, and will generate an alert if
 	// there is a mismatch between .status.currentLinkTarget and
--- a/pkg/common/pv_link_cache.go
+++ b/pkg/common/pv_link_cache.go
@@ -29,7 +29,8 @@ type LocalVolumeDeviceLinkCache struct {
 	client client.Client
 	mgr    manager.Manager
 
-	synced chan struct{}
+	synced        chan struct{}
+	localNodeName string
 	// map of symlinkName in /dev/disk/byid and CurrentBlockDeviceInfo
 	localDeviceInfos map[string]CurrentBlockDeviceInfo
 }
@@ -113,11 +114,12 @@ func (c CurrentBlockDeviceInfo) getLVDLAndPV(ctx context.Context, client client.
 	return lvdl, pv, nil
 }
 
-func NewLocalVolumeDeviceLinkCache(client client.Client, mgr manager.Manager) *LocalVolumeDeviceLinkCache {
+func NewLocalVolumeDeviceLinkCache(client client.Client, mgr manager.Manager, localNodeName string) *LocalVolumeDeviceLinkCache {
 	return &LocalVolumeDeviceLinkCache{
 		client:           client,
 		mgr:              mgr,
 		synced:           make(chan struct{}),
+		localNodeName:    localNodeName,
 		localDeviceInfos: map[string]CurrentBlockDeviceInfo{},
 	}
 }
@@ -138,6 +140,10 @@ func (l *LocalVolumeDeviceLinkCache) IsSynced() bool {
 // caches are started, but we still need to wait for the LVDL informer
 // (which may have been dynamically added) to complete its initial sync.
 func (l *LocalVolumeDeviceLinkCache) Start(ctx context.Context) error {
+	if l.localNodeName == "" {
+		return fmt.Errorf("LocalVolumeDeviceLink cache requires a local node name")
+	}
+
 	informer, err := l.mgr.GetCache().GetInformer(ctx, &v1.LocalVolumeDeviceLink{})
 	if err != nil {
 		return err
@@ -250,6 +256,10 @@ func (l *LocalVolumeDeviceLinkCache) MarkSyncedForTests() {
 }
 
 func (l *LocalVolumeDeviceLinkCache) addOrUpdateLVDL(lvdl *v1.LocalVolumeDeviceLink) {
+	if !l.shouldIndexLVDL(lvdl) {
+		return
+	}
+
 	l.mu.Lock()
 	defer l.mu.Unlock()
 
@@ -287,6 +297,13 @@ func (l *LocalVolumeDeviceLinkCache) addOrUpdateLVDL(lvdl *v1.LocalVolumeDeviceL
 		deviceInfo.lvdls[lvdl.Name] = lvdl
 		l.localDeviceInfos[linkTarget] = deviceInfo
 	}
+}
+
+func (l *LocalVolumeDeviceLinkCache) shouldIndexLVDL(lvdl *v1.LocalVolumeDeviceLink) bool {
+	if lvdl == nil {
+		return false
+	}
+	return lvdl.Spec.NodeName == l.localNodeName
 }
 
 func (l *LocalVolumeDeviceLinkCache) removeLVDL(lvdl *v1.LocalVolumeDeviceLink) {
--- a/test/e2e/gomega_util.go
+++ b/test/e2e/gomega_util.go
@@ -89,43 +89,30 @@ func eventuallyDelete(t *testing.T, removeFinalizers bool, objs ...client.Object
 // using eventuallyDelete is inherently racy
 // this function compares if PV was recreated with a different UID
 // then PV must be deleted.
-func eventuallyDeletePV(t *testing.T, objs ...client.Object) {
+func eventuallyDeletePV(t *testing.T, pv *corev1.PersistentVolume) {
 	f := framework.Global
 	matcher := gomega.NewWithT(t)
-	for _, obj := range objs {
-		accessor, err := meta.Accessor(obj)
-		if err != nil {
-			t.Fatalf("deletion failed, cannot get accessor for object: %+v, obj: %+v", err, obj)
+	oldUID := pv.UID
+	matcher.Eventually(func() error {
+		t.Logf("deleting obj %q with kind %q in ns %q", pv.Name, pv.Kind, pv.Namespace)
+		err := f.Client.Get(context.TODO(), types.NamespacedName{Name: pv.Name, Namespace: pv.Namespace}, pv)
+		if errors.IsNotFound(err) || errors.IsGone(err) {
+			t.Logf("object already deleted: %s", err)
+			return nil
 		}
-		kind := obj.GetObjectKind().GroupVersionKind().Kind
-		name := accessor.GetName()
-		namespace := accessor.GetNamespace()
-		oldUID := accessor.GetUID()
-		matcher.Eventually(func() error {
-			t.Logf("deleting obj %q with kind %q in ns %q", name, kind, namespace)
-			err := f.Client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, obj)
-			if errors.IsNotFound(err) || errors.IsGone(err) {
-				t.Logf("object already deleted: %s", err)
-				return nil
-			}
-			accessor, err = meta.Accessor(obj)
-			if err != nil {
-				t.Fatalf("deletion failed, cannot get accessor for object: %+v, obj: %+v", err, obj)
-			}
-			newUID := accessor.GetUID()
-			if newUID != oldUID {
-				t.Logf("object %s has been deleted and recreated", name)
-				return nil
-			}
+		newUID := pv.GetUID()
+		if newUID != oldUID {
+			t.Logf("object %s has been deleted and recreated", pv.Name)
+			return nil
+		}
 
-			err = f.Client.Delete(context.TODO(), obj)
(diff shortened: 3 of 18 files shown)


pkg/common/pv_link_cache.go
```diff
@@ -282,6 +300,13 @@ func (l *LocalVolumeDeviceLinkCache) addOrUpdateLVDL(lvdl *v1.LocalVolumeDeviceL
 	}
 }
 
+func (l *LocalVolumeDeviceLinkCache) shouldIndexLVDL(lvdl *v1.LocalVolumeDeviceLink) bool {
+	if lvdl == nil {
+		return false
+	}
+	return lvdl.Spec.NodeName == l.localNodeName
```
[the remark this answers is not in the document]
gnufied: This is fair assessment but we are yet to ship 4.22 version of LSO, which will have LVDL objects first time and hence upgrade concerns are not really an issue.

api/v1/localvolumedevicelink_types.go
```diff
@@ -52,6 +52,13 @@ type LocalVolumeDeviceLinkSpec struct {
 	// +kubebuilder:validation:MaxLength=253
 	PersistentVolumeName string `json:"persistentVolumeName"`
 
+	// nodeName is the name of the Openshift node on which this
```
jsafrane: nit: 
```suggestion
	// nodeName is the name of the OpenShift node on which this
```

api/v1/localvolumedevicelink_types.go
```diff
@@ -52,6 +52,13 @@ type LocalVolumeDeviceLinkSpec struct {
...
+	// LocalVolumeDeviceLink object exists
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	NodeName string `json:"nodeName"`
```
everettraven: If the empty string is not a valid input (based on the min length of 1 constraint it is not), add `omitempty` to prevent this field from being serialized as `nodeName: ""` by structured clients.

api/v1/localvolumedevicelink_types.go
everettraven: Node names look to be validated as RFC1123 subdomain names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

You can add a CEL validation like https://github.com/openshift/api/blob/3c6b218b8a80015f52413591b133fa6391d6ba1a/machineconfiguration/v1/types_machineosconfig.go#L177 to enforce that format.

api/v1/localvolumedevicelink_types.go
```diff
@@ -52,6 +52,13 @@ type LocalVolumeDeviceLinkSpec struct {
 	// +kubebuilder:validation:MaxLength=253
 	PersistentVolumeName string `json:"persistentVolumeName"`
 
+	// nodeName is the name of the Openshift node on which this
+	// LocalVolumeDeviceLink object exists
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
everettraven, line 59: Make sure to document these constraints in the GoDoc, otherwise users will be unaware of them until they run into them or intentionally inspect the CRD schema.

api/v1/localvolumedevicelink_types.go
gnufied: ack. added.

api/v1/localvolumedevicelink_types.go
gnufied: fixed.

gnufied: So, the validation IMO makes sense if it were a standalone field but in this case this is another node's name. I am not sure if we need to validate it separately with CEL.  Node names are already validated by api-server. Yes, we accept here any string but if user specifies invalid nodename then it is on them if it doesn't matches any real node. Moreover, this is anyways created by LSO itself.

I think there is a cost to all these validations and I am happy to copy and paste, but IMO CEL validation here doesn't add any value.

test/e2e/localvolume_test.go
```diff
@@ -356,29 +438,85 @@ func verifyProvisionerAnnotation(t *testing.T, pvs []corev1.PersistentVolume, no
...
 	}, time.Minute*3, time.Second*2).ShouldNot(gomega.HaveOccurred(), "cleaning up pvs for lv: %q", localVolume.GetName())
 
+	// Delete the owning resources after their PVs are fully gone so the
+	// diskmaker cleanup path can remove PV finalizers first.
+	eventuallyDelete(t, true, localVolume)
+	sc := &storagev1.StorageClass{
+		TypeMeta:   metav1.TypeMeta{Kind: localv1.LocalVolumeKind},
+		ObjectMeta: metav1.ObjectMeta{Name: localVolume.Spec.StorageClassDevices[0].StorageClassName},
+	}
+	eventuallyDelete(t, false, sc)
```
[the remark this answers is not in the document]
gnufied: I have replaced this with `eventuallyDelete` which should actually verify PV to be really deleted.

test/e2e/localvolumeset_test.go
```diff
@@ -151,6 +153,85 @@ func LocalVolumeSetTest(ctx *framework.TestCtx, cleanupFuncs *[]cleanupFn) func(
...
+			sharedPVNames = append(sharedPVNames, pv.Name)
+		}
+		sharedLVDLs = eventuallyFindLVDLsForPVs(t, f, namespace, sharedPVNames)
+		assertLVDLsContainTargetAndNodes(t, sharedLVDLs, sharedScsi8Link, []string{nodeEnv[0].node.Name, nodeEnv[1].node.Name})
+
+		t.Log("cleaning up LocalVolumeSet duplicate by-id reproducer before continuing with standard test flow")
+		eventuallyDelete(t, false, sharedLVSet)
+		waitForLVSetAndOwnedPVsToDisappear(t, sharedLVSet)
+		removeUdevSymlink(t, ctx, nodeEnv[0].node.Labels[corev1.LabelHostname], sharedScsi8Link)
+		removeUdevSymlink(t, ctx, nodeEnv[1].node.Labels[corev1.LabelHostname], sharedScsi8Link)
```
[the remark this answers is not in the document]
gnufied: SC is automatically removed when LV and LVSet is deleted (without removing finalizer anyways). It should not be necessary to remove Storageclass explicitly.

test/e2e/localvolumeset_test.go
```diff
@@ -504,30 +585,73 @@ func LocalVolumeSetTest(ctx *framework.TestCtx, cleanupFuncs *[]cleanupFn) func(
...
-					eventuallyDelete(t, false, &pv)
-				}
+				eventuallyDeletePV(t, &pv)
 			}
 			return nil
 		}, time.Minute*3, time.Second*2).ShouldNot(gomega.HaveOccurred(), "cleaning up pvs for lvset: %q", lvset.GetName())
+
+		sc := &storagev1.StorageClass{ObjectMeta: metav1.ObjectMeta{Name: lvset.Spec.StorageClassName}}
+		eventuallyDelete(t, true, lvset)
+		eventuallyDelete(t, false, sc)
```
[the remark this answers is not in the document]
gnufied: fixed it by replacing it with `eventuallyDelete`

test/e2e/symlink_check.go
```diff
@@ -279,7 +279,8 @@ func verifyMultiStepPreferredLinkReconciliation(
 		t.Logf("step %d: deleting PV %q and verifying recreation", i+1, pv.Name)
 		oldPVUID := pv.UID
 		oldPVPath := pv.Spec.Local.Path
-		eventuallyDelete(t, false, &pv)
+		err := f.Client.Delete(goctx.TODO(), &pv)
+		matcher.Expect(err).Error().To(gomega.BeNil(), "expected no error after pv delete")
```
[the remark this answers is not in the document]
gnufied: fixed in new push.

api/v1/localvolumedevicelink_types.go
gnufied: I am not sure if we publish any godoc directly from this repo. But sure, will keep in mind.

test/e2e/gomega_util.go
```diff
@@ -85,6 +85,49 @@ func eventuallyDelete(t *testing.T, removeFinalizers bool, objs ...client.Object
 
 }
 
+// PVs rapidly get deleted and recreated if diskmaker is running
+// using eventuallyDelete is inherently racy
+// this function compares if PV was recreated with a different UID
+// then PV must be deleted.
+func eventuallyDeletePV(t *testing.T, objs ...client.Object) {
```
jsafrane: * Why `client.Object`? I used `*corev1.PersistentVolume`  and all callers are fine with that. It will dramatically simplify the function.
* Why variable nr. of arguments? All callers use a single parameter.

test/e2e/localvolume_test.go
```diff
@@ -110,6 +110,85 @@ func LocalVolumeTest(ctx *framework.Context, cleanupFuncs *[]cleanupFn) func(*te
...
+				nodeHostname: nodeEnv[1].node.Labels[corev1.LabelHostname],
+				currentLink:  currentSymlinkForDisk(sharedAliasDisk),
+			},
+		}, sharedScsi8Link)
+
+		sharedLocalVolume := getLocalVolume(selectedNode, sharedScsi8Link, namespace)
+		sharedLocalVolume.Name = "test-local-disk-shared-byid"
+		sharedLocalVolume.Spec.StorageClassDevices[0].StorageClassName = "test-local-sc-shared-byid"
+
+		matcher.Eventually(func() error {
```
jsafrane: e2e tests usually don't retry API calls, we expect the API server is stable.

test/e2e/localvolume_test.go
```diff
@@ -356,29 +435,77 @@ func verifyProvisionerAnnotation(t *testing.T, pvs []corev1.PersistentVolume, no
...
+				pvNames = append(pvNames, pv.Name)
+			}
+			return fmt.Errorf("waiting for owned PVs to disappear for localvolume %q: %+v", localVolume.Name, pvNames)
+		}
+
+		return nil
+	}, time.Minute*8, time.Second*5).ShouldNot(gomega.HaveOccurred(), "waiting for localvolume %q and owned PVs to be deleted", localVolume.Name)
+}
+
+func cleanupLVAndWaitForOwnedPVsToDisappear(t *testing.T, f *framework.Framework, localVolume *localv1.LocalVolume) {
```
jsafrane, line 474: This is almost the same as `cleanupLVResources`, which applies certain amount of force (deletes PVs, removes LV finalizers). Why do we need to delete PVs and remove finalizers in the first place? If it's needed, a comment why we need both function + a better name (`forceCleanupLVResources`?) would be appropriate.

test/e2e/localvolumeset_test.go
```diff
@@ -151,6 +153,85 @@ func LocalVolumeSetTest(ctx *framework.TestCtx, cleanupFuncs *[]cleanupFn) func(
...
+		sharedLVDLs := eventuallyFindLVDLsForPVs(t, f, namespace, sharedPVNames)
+		assertLVDLsContainTargetAndNodes(t, sharedLVDLs, sharedScsi8Link, []string{nodeEnv[0].node.Name})
+
+		matcher.Eventually(func() error {
+			t.Log("expanding shared localvolumeset reproducer to a second node")
+			key := types.NamespacedName{Name: sharedLVSet.GetName(), Namespace: sharedLVSet.GetNamespace()}
+			if err := f.Client.Get(context.TODO(), key, sharedLVSet); err != nil {
+				return err
+			}
+			sharedLVSet.Spec.MaxDeviceCount = &two
```
jsafrane: Is the `MaxDeviceCount` update necessary? IIRC it's nr. of devices per node, not per the whole LVSet.

test/e2e/localvolume_test.go
gnufied: Yes, it should be sufficient to delete LV (or LVSet) and then just wait for PVs, SCs everything to disappear automatically. There should not be any need to delete PVs or remove finalizers manually. 

Having said that, I spent hours debugging issues around cleanup (and even found a race condition which happens if you delete PV and then LVSet manually - https://redhat.atlassian.net/browse/OCPBUGS-83626 ), so I did not wanted to touch old cleanup functions. 

There is lot of room for improvement in cleanup routines.

api/v1/localvolumedevicelink_types.go
everettraven: The cost is pretty minimal for an improved user experience.

If you feel strongly that users should be allowed to provide an input that doesn't match the format of a valid node name then I won't stop you.

It is normally easier for end-users to get earlier feedback (through an admission failure) than have to dig to find out that there is some status condition representing a failure state because the node name doesn't follow the correct format but if you think that will be a rare occasion :shrug:.

api/v1/localvolumedevicelink_types.go
everettraven: If this is a CRD, the OpenAPI description for the field is generated from this GoDoc and will not include the markers.

Users using `kubectl explain ...` (or anything else that pulls only the description from the schema) to learn about the API surface will not be aware of these constraints.

api/v1/localvolumedevicelink_types.go
gnufied: My concern is - adding the CEL validation adds 0 value to this API because of two reasons:

1. This is an existing node name and hence has already went through validation.
2. This API is actually not end-user facing for the most part. These objects are created by LSO automatically. Users will never be setting `nodeName` themselves (the only field they can set themselves is `policy` field)

The downside is - even if there is pretty minimal overhead of applying the CEL validation, there is an overhead in terms of maintaining this CEL validation. While I can copy/paste the validation, I do not understand it very well. If a version of api-server breaks it, we will be dealing with the fallout having copied the code we did not fully understand. 

If the validation prevented a real-world issue then it is worth keeping it and dealing with associated overhead but IMO it doesn't seem like worth it here.

test/e2e/gomega_util.go
gnufied: fixed.

test/e2e/localvolumeset_test.go
gnufied: fixed.

test/e2e/localvolume_test.go
gnufied: On the other hand, I do need this new way of cleaning up stuff without explicitly removing finalizers etc, because calling `cleanupLVResources` in the middle of the testsuite appears to be triggering the race condition I mentioned above.

I have filed https://redhat.atlassian.net/browse/OCPBUGS-83729 to cleanup the e2e some more.

jsafrane:
/lgtm

I tested both LocalVolume and LocalVolumeSet on a cluster with 3 workers. Each worker had 1 extra disk, with manually created `/dev/disk/by-id/wwn-jsafrane` -> `/dev/nvme1n1`. LSO created 3 PVs and a StatefulSet used them via PVCs. The PVs survived reboot, upgrade to RHEL10 (with LV, I did not test LVSet) and PV re-creation (i.e. StatefulSet / PVC re-creation).
There is new e2e test to prevent regressions both in LocalVolume and LocalVolumeSet.
/verified by jsafrane

PR #614, opened 2026-04-22
OCPBUGS-84123,STOR-2682: Add e2e test for LSO symlink alerts

jsafrane:
We promised this e2e test in the [KEP](https://github.com/openshift/enhancements/blob/master/enhancements/storage/lso-recreate-symlinks.md#test-plan).

It checks both Localvolumes and LocalVolumeSets. It does not wait for the alert to get Firing, that would take 15 minutes. Pending alert must be enough..

And fix a bug revealed by the test - each diskmaker should report only problems of LocalVolumeDevice links managed by the diskmaker and ignore all others. Each diskmaker reporting each LVDL messed up the e2e tests.

Unfortunately, due to Prometheus scraping interval (60s), the new checks add quite some delay to the job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Device-link metrics are now scoped to the local node, improving per-node accuracy for mismatch and missing-path metrics.

* **Tests**
  * Added Prometheus helpers and polling utilities for e2e metric assertions.
  * e2e flows now wait for specific metric and alert states during multi-step reconciliation.
  * Added tests to ensure links from other nodes are ignored and a pre-deploy step to enable namespace metrics.

* **Chores**
  * Updated module dependency metadata.


test/e2e/main_test.go
```diff
@@ -146,3 +152,35 @@ func waitForOperatorToBeReady(t *testing.T, ctx *framework.TestCtx) error {
...
+// The namespace created by the e2e job often doesn't have it enabled.
+// Enable it now.
+func enableNamespaceMetrics(t *testing.T, ctx *framework.TestCtx) error {
+	namespace, err := ctx.GetOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	f := framework.Global
+
+	ns, err := f.KubeClient.CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
```
gnufied: nit: `t.Context()`

test/e2e/main_test.go
```diff
@@ -146,3 +152,35 @@ func waitForOperatorToBeReady(t *testing.T, ctx *framework.TestCtx) error {
...
+		ns.Labels = make(map[string]string)
+	}
+	if ns.Labels["openshift.io/cluster-monitoring"] == "true" {
+		t.Logf("Namespace %s already has cluster-monitoring enabled", namespace)
+		return nil
+	}
+
+	t.Logf("Enabling cluster-monitoring for namespace %s", namespace)
+	ns.Labels["openshift.io/cluster-monitoring"] = "true"
+	_, err = f.KubeClient.CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{})
```
gnufied, line 180: same as above.

test/e2e/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+		return result[0].Value
+	}, time.Minute*5, time.Second*5).Should(gomega.Equal(value))
+}
+
+// Copied & updated from openshift/origin test/extended/util/prometheus/helpers.go
+func runQueryAtTime(t *testing.T, prometheusClient prometheusv1.API, query string, evaluationTime time.Time) (model.Vector, error) {
+	var lastErr error
+	var result model.Value
+	var warnings prometheusv1.Warnings
+	for i := 0; i < 5; i++ {
```
gnufied: replace with - 

```
for i := range 5 {
```

test/e2e/symlink_check.go
```diff
@@ -302,7 +319,14 @@ func verifyMultiStepPreferredLinkReconciliation(
 			"step %d: LVDL PreferredLinkTarget after PV recreation", i+1)
 		matcher.Expect(lvdl.Spec.Policy).To(gomega.Equal(localv1.DeviceLinkPolicyPreferredLinkTarget),
 			"step %d: LVDL policy must remain PreferredLinkTarget after PV recreation", i+1)
+		waitForMetric(t, prometheusClient, fmt.Sprintf("lso_device_link_mismatch{name=\"%s\", policy=\"None\"}", lvdl.Name), 0)
+		waitForMetric(t, prometheusClient, fmt.Sprintf("lso_device_link_mismatch{name=\"%s\", policy=\"CurrentLinkTarget\"}", lvdl.Name), 0)
```
gnufied: it might be easier to just set:

```
waitForMetric(t, prometheusClient, fmt.Sprintf("lso_device_link_mismatch{name=\"%s\"}", lvdl.Name), 0)
```

jsafrane: Good point!

PR #615, opened 2026-04-24
OCPBUGS-84330: Ensure lvdl update

gnufied:
Fixes https://redhat.atlassian.net/browse/OCPBUGS-84330

## Summary by CodeRabbit

* **Bug Fixes**
  * Symlink status recomputation is recorded only when meaningfully changed and cache updates occur only after successful API persistence; preferred-link changes now emit explicit change events.

* **New Features**
  * New typed error for non-preferred-policy cases exposes the affected device-link to callers.
  * Provisioning flow replaced with a reusable syncer and reconciler initialization now surfaces startup errors.

* **Refactor**
  * Controllers share a single device-link handler and centralized update path.

* **Tests**
  * Tests extended for status recomputation, preferred-link updates, and constructor validation.

* **Documentation**
  * Added agent guidance doc and a pointer file.


pkg/common/device_link_handler.go
```diff
@@ -163,6 +163,33 @@ func (dl *DeviceLinkHandler) ApplyStatus(ctx context.Context, pvName, namespace
...
+	if existing.Status.PreferredLinkTarget != copyToUpdate.Status.PreferredLinkTarget {
+		infoUpdate := fmt.Sprintf("PreferredLinkTarget has changed from %s to %s for device %s", existing.Status.PreferredLinkTarget, copyToUpdate.Status.PreferredLinkTarget, blockDevice.Name)
+		klog.Info(infoUpdate)
+		dl.recorder.Eventf(ownerObj, corev1.EventTypeNormal, "PreferredSymlinkChanged", infoUpdate)
+	}
+
+	err = dl.client.Status().Update(ctx, copyToUpdate)
+	if err == nil {
+		dl.cacheWriter.AddOrUpdateLVDL(copyToUpdate)
+	}
```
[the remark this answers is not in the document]
gnufied: That is okay. See the `PreferredLinkTarget` is mainly telling users that whatever device their PV is using has changed its symlink. It does not have to be necessarily recorded in the API at this point.

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -480,6 +481,69 @@ func (r *LocalVolumeReconciler) resolveValidDeviceLocation(devicePath string, fo
...
+		}
+
+		if lvdl != nil {
+			// if we are here that means, lvdl exists but user has not chosen to use PreferredLinkTarget policy
+			// and user provided path does not exist
+			_, err := r.deviceLinkHandler.UpdateDeviceLinks(ctx, lvdl, deviceLocation.BlockDevice, deviceLocation.UserProvidedPath)
+			if err != nil {
+				return deviceLocation, fmt.Errorf("user provided path %s does not exist for LocalVolume %s and failed to updated lvdl %v", deviceLocation.UserProvidedPath, r.localVolume.Name, err)
+			}
+		}
```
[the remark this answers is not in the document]
gnufied: This comment is on outdated diff.

pkg/common/device_link_handler.go
```diff
@@ -40,14 +40,17 @@ type DeviceLinkHandler struct {
...
+		return nil, fmt.Errorf("node name is required for DeviceLinkHandler")
+	}
 	return &DeviceLinkHandler{
 		client:       client,
 		clientReader: clientReader,
 		recorder:     recorder,
 		cacheWriter:  cacheWriter,
 		nodeName:     nodeName,
-	}
+	}, nil
```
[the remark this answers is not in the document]
gnufied: `cacheWriter` can't be nil. It has been properly initialized when being set.

pkg/common/device_link_handler.go
```diff
@@ -163,6 +163,33 @@ func (dl *DeviceLinkHandler) ApplyStatus(ctx context.Context, pvName, namespace
...
 
+	if existing.Status.PreferredLinkTarget != copyToUpdate.Status.PreferredLinkTarget {
+		infoUpdate := fmt.Sprintf("PreferredLinkTarget has changed from %s to %s for device %s", existing.Status.PreferredLinkTarget, copyToUpdate.Status.PreferredLinkTarget, blockDevice.Name)
+		klog.Info(infoUpdate)
+		dl.recorder.Eventf(ownerObj, corev1.EventTypeNormal, "PreferredSymlinkChanged", infoUpdate)
+	}
+
+	err = dl.client.Status().Update(ctx, copyToUpdate)
+	if err == nil {
+		dl.cacheWriter.AddOrUpdateLVDL(copyToUpdate)
```
jsafrane: Why copying the code? Call `UpdateDeviceLinks` instead.

pkg/common/provisioner_utils.go
```diff
@@ -65,9 +65,19 @@ type CreateLocalPVArgs struct {
...
 }
 
-// CreateLocalPV is used to create a local PV against a symlink
-// after passing the same validations against that symlink that local-static-provisioner uses
-func CreateLocalPV(ctx context.Context, args CreateLocalPVArgs) error {
+// PVAndLVDLSyncer synchronizes the PV and LocalVolumeDeviceLink state for a symlinked device.
+type PVAndLVDLSyncer struct {
+	args PVAndLVDLSyncArgs
+}
+
```
jsafrane: I find this struct very unnecessary. What value does it add?

pkg/diskmaker/controllers/lv/reconcile.go
```diff
@@ -621,7 +630,7 @@ func (r *LocalVolumeReconciler) provisionPV(ctx context.Context, scName string,
 		CacheWriter:           r.pvLinkCache,
 	}
 
-	return common.CreateLocalPV(ctx, createLocalPVArgs)
+	return common.NewPVAndLVDLSyncer(syncArgs).Sync(ctx)
```
jsafrane: I don't see any improvement over the original function call, only unnecessary complexity.

gnufied: Well, the thinking is - this expresses the intent better. We are not merely creating Local PVs, but we are also syncing LVDL objects. We could also rename this simply to `common.SyncLVDLAndPV` and remove the struct.

pkg/common/device_link_handler.go
gnufied: removed.

pkg/common/provisioner_utils.go
gnufied: removed.

pkg/diskmaker/controllers/lv/reconcile.go
gnufied: I just renamed the function and removed the code around struct etc.

gnufied:
The failure looks unrelated:

```
  failed to wait for importing imagestreamtags on ci-op-1d71bi31/stable: failed to reimport the tag ci-
op-1d71bi31/stable:vcf-migration-operator: unable to import tag ci-op-1d71bi31/stable:vcf-migration-operator 
with message Internal error occurred: [dockerimage.image.openshift.io "quay.io/openshift/ci:ocp_5.0_vcf-
migration-operator" not found, dockerimage.image.openshift.io "quay-proxy.ci.openshift.org/openshift/
ci:ocp_5.0_vcf-migration-operator" not found] on the image stream even after (6) imports: timed out waiting for 
the condition}
```

I am checking with splat team about what is going on.

pkg/diskmaker/controllers/lv/create_pv_test.go
```diff
@@ -308,7 +308,7 @@ func TestCreatePV(t *testing.T) {
 //   - The symlink matcher would receive DevicePath and filepath.Base of that
 //     would still be "sda", so to make the test more robust the mock checks
 //     the exact string passed to FilePathEvalSymLinks as its target.
-func TestCreateLocalPV_DeviceLinkArgOrder(t *testing.T) {
+func TestSyncLVAndLVDL_DeviceLinkArgOrder(t *testing.T) {
```
jsafrane: `TestSyncSyncPVAndLVDL_*`, applies also to the comments above and below.

pkg/diskmaker/controllers/lv/create_pv_test.go
```diff
@@ -406,7 +406,7 @@ func TestCreateLocalPV_DeviceLinkArgOrder(t *testing.T) {
 		"FilePathEvalSymLinks should have been called with the by-id link")
 }
 
-func TestCreateLocalPV_DeviceLinkLifecycle(t *testing.T) {
+func TestSyncLVAndLVDL_DeviceLinkLifecycle(t *testing.T) {
```
jsafrane: `TestSyncSyncPVAndLVDL_*`, applies also to the comments above and below.

pkg/common/current_block_device_info.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return nil, "", false
+}
+
+type CurrentBlockDeviceInfo struct {
+	// map of lvdlname and LocalVolumeDeviceLink
+	lvdls map[string]*v1.LocalVolumeDeviceLink
+}
+
+// GetPVSymlinkPath returns a symlinkPath in /mnt/local-storage for Local volumes.
```
jsafrane:
```suggestion
// GetPVSymlinkPath finds a symlinkPath in /mnt/local-storage from
// LVDL for a given newSymlinkSourcePath. If the symlinkPath needs
// a change and the LVDL policy is not to update existing symlinks,
// then PolicyNotPreferredError error is returned.
```

pkg/common/current_block_device_info.go
```diff
@@ -0,0 +1,135 @@
...
+// Arguments:
+// symlinkDir is path in /mnt/local-storage with storageclass name:
+//
+//	example - /mnt/local-storage/foobar/
+//
+// newSymlinkSourcePath is path in /dev/disk/by-id which points to current device:
+//
+//	example - /dev/disk/by-id/wwn-0x123432
+//
+// Assuming existing currentLinkTarget somehow doesn't resolve, we could be in this code path.
```
jsafrane:
```suggestion
// This function MUST be called only when there is no known
// `/mnt/local-storage/<sc>/symlink` for `newSymlinkSourcePath`.
// and this function is the last resort to find it.
```

If called with a valid `/mnt/local/<sc>symlink`, it may throw `PolicyNotPreferredError`.

pkg/common/current_block_device_info.go
```diff
@@ -0,0 +1,135 @@
...
+		// A single device is used by multiple objects with different StorageClasses (= different symlinkDirs). That should never happen.
+		return "", fmt.Errorf("discovered symlink directory %s does not match expected symlink directory %s for device %s - is this device matched by multiple LocalVolumes / LocalVolumeSets?", discoveredSymlinkDir, symlinkDir, newSymlinkSourcePath)
+	}
+	currentLinkTarget := lvdl.Status.CurrentLinkTarget
+	validLinkTargets := lvdl.Status.ValidLinkTargets
+	policy := lvdl.Spec.Policy
+
+	if policy != v1.DeviceLinkPolicyPreferredLinkTarget {
+		return "", PolicyNotPreferredError{SymlinkSource: newSymlinkSourcePath, SymlinkDir: symlinkDir, LVDL: lvdl, PVSymlinkPath: symlinkPath}
+	}
```
jsafrane, line 93: Here I guess we will agree to disagree again. There is no error in `GetPVSymlinkPath` at this point, yet an error is returned. The `PolicyNotPreferredError` is used to hack different behavior to all `GetPVSymlinkPath` callers via error path.

Fixing that would be a huge refactoring and I don't insist on it, but I personally think it makes the code more confusing.

jsafrane (commented):
It looks much better now, thanks for the renames and comment.

I have a few suggestions how to make some descriptions more clear, nothing too serious.

pkg/common/current_block_device_info.go
gnufied: hmm, so just for posterity - here is how I see it.

We know that `GetPVSymlinkPath` will *never* get called if things were working as expected. The fact that `GetPVSymlinkPath` is called means - the symlink path in `/mnt/local-storage` is **broken**.  It basically means PV is broken.  We have to realize that, If `GetPVSymlinkPath` is being called, that means - we **are already in error** state. Question is - can we recover from it? 

So `GetPVSymlinkPath` is our hail mary attempt to find a valid path for current device, if we find it but we can't use it, because policy on LVDL prevents us from using it - then is that not an error?

gnufied: I am tempted to rename `GetPVSymlinkPath` one more time to `RecoverPVSymlinkPath`, because that is what we we doing. It is not an ordinary `GetPVSymlinkPath`, it is our *last* resort. If we can't get the path here, that means associated PV is broken until user manually fixes it or changes the policy.

pkg/diskmaker/controllers/lv/create_pv_test.go
gnufied: fixed.

pkg/diskmaker/controllers/lv/create_pv_test.go
gnufied: fixed.

pkg/common/current_block_device_info.go
gnufied: fixed.

pkg/common/current_block_device_info.go
gnufied: I incorporated this change, but renamed  "no known" to "no valid". There may still be symlinks in `/mnt/local-storage` for this device - but they are likely not resolving anymore..

pkg/common/current_block_device_info.go
jsafrane: > I am tempted to rename GetPVSymlinkPath one more time to RecoverPVSymlinkPath

That's actually a very good name for the function.

PR #618, opened 2026-04-29
OCPBUGS-83305: record symlinkPath in LVDL

jsafrane:
For cases where PV does not exist, for example during PV re-creation after PVC was deleted, the diskmaker needs a way to discover `symlinkPath` of a device targeted by LocalVolumeSet to be able to re-create the PV. The current way of discovering the `symlinkPath` from PV won't work in this case.

Add a new field to LocalVolumeDeviceLink `status`. It is immutable, because the same `symlinkPath` is already used in PV `spec.local.path`, which is immutable already.

And since it's immutable, it does not really matter if the diskmaker loads the value from the PV first and if it's missing, them from LVDL status or the other way around - both ways should return the same result. (Unless the field has been just introduced in LVDL and is empty.) So let's read it from LVDL first and fall back to PV only if it's missing there.

## Summary by CodeRabbit

* **New Features**
  * Added a new optional status field to record the PersistentVolume symlink path (string, max 4096, immutable once set).

* **Behavior Changes**
  * Status symlink path is now preferred when resolving symlink targets and is persisted in status; resolution more explicitly handles missing/invalid PVs.

* **Tests**
  * Unit and e2e tests expanded to validate propagation, precedence, immutability, and error scenarios for the symlink path.

### DIFF
--- a/api/v1/localvolumedevicelink_types.go
+++ b/api/v1/localvolumedevicelink_types.go
@@ -95,6 +95,16 @@ type LocalVolumeDeviceLinkStatus struct {
 	// +listType=set
 	// +kubebuilder:validation:MaxItems=256
 	ValidLinkTargets []string `json:"validLinkTargets,omitempty"`
+	// persistentVolumeSymlinkPath is the symlink in /mnt/local-storage directory that points to
+	// the actual local device.
+	// This path is used by corresponding PersistentVolume (PV) object for all operations (e.g. mount).
+	// LSO creates and updates this symlink for the whole lifetime of the PV.
+	// This field is immutable once set, because the path in the corresponding
+	// PV object is immutable too.
+	// +optional
+	// +kubebuilder:validation:MaxLength=4096
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf || oldSelf == ''",message="persistentVolumeSymlinkPath is immutable once set"
+	PersistentVolumeSymlinkPath string `json:"persistentVolumeSymlinkPath,omitempty"`
 	// filesystemUUID is the UUID of the filesystem found on the device (when available)
 	// +optional
 	// +kubebuilder:validation:MinLength=1
--- a/pkg/common/pv_link_cache.go
+++ b/pkg/common/pv_link_cache.go
@@ -70,7 +70,7 @@ func (c CurrentBlockDeviceInfo) GetSymlinkTargetPath(ctx context.Context, symlin
 		return "", fmt.Errorf("unexpected empty lvdl set for symlink %s", newSymlinkSourcePath)
 	}
 
-	lvdl, pv, err := c.getLVDLAndPV(ctx, client)
+	lvdl, symlinkPath, err := c.getLVDLAndSymlinkPath(ctx, client)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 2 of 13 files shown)


api/v1/localvolumedevicelink_types.go
```diff
@@ -95,6 +95,15 @@ type LocalVolumeDeviceLinkStatus struct {
...
 	ValidLinkTargets []string `json:"validLinkTargets,omitempty"`
+	// symlinkPath is the path of the symlink in /mnt/local-storage directory that points to
+	// the actual local device.
+	// This path is used by corresponding PersistentVolume (PV) object for all operations (e.g. mount).
+	// LSO creates and updates this symlink for the whole lifetime of the PV.
+	// The symlink path is immutable once set.
+	// +optional
+	// +kubebuilder:validation:MaxLength=4096
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf || oldSelf == ''",message="symlinkPath is immutable once set"
+	SymlinkPath string `json:"symlinkPath,omitempty"`
```
gnufied: I wonder if we should call it `PVSymlinkPath` to be more accurate. This will be pretty clear then that this is the path used by PVs.

pkg/common/pv_link_cache.go
```diff
@@ -101,17 +97,39 @@ func (c CurrentBlockDeviceInfo) GetSymlinkTargetPath(ctx context.Context, symlin
 	return filepath.Join(symlinkDir, symlinkBaseName), nil
```
gnufied: we could just return `symlinkPath` returned by `getLVDLAndSymlinkPath` ? and in which case `GetSymlinkTargetPath` probably doesn't need `symlinkDir` argument.

api/v1/localvolumedevicelink_types.go
jsafrane: No abbrevs. in the API, so `PersistentVolumeSymlinkPath`. I can live with it.
(which reminds me, I need to update the enhancement)

gnufied: `PersistentVolumeSymlinkPath` is fine too.

jsafrane: Renamed to PersistentVolumeSymlinkPath

pkg/common/pv_link_cache.go
gnufied: ping on this..

pkg/common/pv_link_cache.go
```diff
@@ -88,11 +88,7 @@ func (c CurrentBlockDeviceInfo) GetSymlinkTargetPath(ctx context.Context, symlin
 		return "", fmt.Errorf("currentSymlink %s still resolves to %s for %s", currentLinkTarget, resolvedCurrent, newSymlinkSourcePath)
 	}
 
-	if pv.Spec.Local == nil || pv.Spec.Local.Path == "" {
-		return "", fmt.Errorf("pv %s has empty local path", pv.Name)
-	}
-	currentTargetPath := pv.Spec.Local.Path
-	symlinkBaseName := filepath.Base(currentTargetPath)
+	symlinkBaseName := filepath.Base(symlinkPath)
```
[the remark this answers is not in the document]
jsafrane: Only cluster admin can set such invalid symlinkPath.

pkg/common/pv_link_cache.go
jsafrane: I added a check for unexpected `<sc>` in `/mnt/local-storage/<sc>` to make sure we don't accidentally get a LVDL for a device with a different StorageClass (i.e. managed by a different LV / LVSet). Then yes, we don't need to parse + join the `symlinkPath` and I fixed that.

Whoever, are you saying that the checks for `internal.FilePathEvalSymLinks` and `slices.Contains(validLinkTargets, newSymlinkSourcePath)` above are not needed now and GetSymlinkTargetPath can return whatever getLVDLAndSymlinkPath returns? Why were the checks needed before?

api/v1/localvolumedevicelink_types.go
jsafrane: btw, I thought about renaming `symlinkPath` to `pvSymlinkPath` even in the code, but it would be a huge change. `symlinkPath` is used all over the code, not just in the parts that I modified. I'd prefer to keep it as it is and maybe fix it in a separate PR.

PR #625, opened 2026-05-20
OCPBUGS-79680: Emit correct metrics for orphan volumes.

gnufied:
Fixes https://redhat.atlassian.net/browse/OCPBUGS-79680

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus gauges to monitor orphaned symlinks for LocalVolume and LocalVolumeSet.

* **Improvements**
  * Reconciler now separates spec-matched devices from provisioning-eligible devices for more accurate orphan detection.
  * Default device-filtering behavior renamed/clarified to improve inclusion/exclusion semantics.

* **Tests**
  * New unit tests for orphaned-symlink metric calculation and spec-matching device selection.

### DIFF
--- a/pkg/diskmaker/controllers/lvset/reconcile.go
+++ b/pkg/diskmaker/controllers/lvset/reconcile.go
@@ -307,7 +307,8 @@ func (r *LocalVolumeSetReconciler) Reconcile(ctx context.Context, request ctrl.R
 	// update metrics for total persistent volumes provisioned
 	localmetrics.SetLVSProvisionedPVMetric(nodeName, storageClassName, totalProvisionedPVs)
 
-	orphanSymlinkDevices, err := internal.GetOrphanedSymlinks(symLinkDir, validDevices)
+	specMatchedDevices := getSpecMatchedDevices(lvset, blockDevices)
+	orphanSymlinkDevices, err := internal.GetOrphanedSymlinks(symLinkDir, specMatchedDevices)
 	if err != nil {
 		klog.ErrorS(err, "failed to get orphaned symlink devices in current reconcile")
 	}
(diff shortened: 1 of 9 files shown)


pkg/diskmaker/controllers/lvset/reconcile.go
```diff
@@ -502,6 +504,47 @@ DeviceLoop:
 	return validDevices, delayedDevices, rejectedDevices
 }
 
+// getSpecMatchedDevices returns devices that match the LVSet's inclusion and
+// exclusion spec criteria, ignoring provisioning-eligibility filters (such as
+// filesystem signature, exclusive access, bind mounts). This broader list is
+// used for orphan detection so that already-provisioned devices are not
+// incorrectly counted as orphaned.
+func getSpecMatchedDevices(
```
jsafrane: This function looks quite similar to `getValidDevices()`. `Reconcile()` above has:
https://github.com/openshift/local-storage-operator/blob/81f2fef8e69774ab9e100b0871e0c643c6aee078/pkg/diskmaker/controllers/lvset/reconcile.go#L261-L262

Instead of calling a new function `getSpecMatchedDevices()`, can `Reconcile` just join `validDevices`, `delayedDevices` and `rejectedDevices`?

Some refactoring in `Reconcile()` would be needed to evaluate `matcher` / `excluder` first and then `filter` and deviceAge, otherwise `delayedDevices` and `rejectedDevices` will have devices not matched by the matcher / rejected by `excluder`.

It would even fix a bug / inconvenience in `getValidDevices()` - now it reports "*found possible matching disk, waiting %v to claim*" for a device, but after 1 minute it can change its mind and reject the device, because it was not matched by `matcher` or rejected by `excluder`.

jsafrane: Alternatively, make a function to evaluate `matcherMap` + `excluderMap` on a `BlockDevice` and call it both in `getSpecMatchedDevices` and `getValidDevices()`

jsafrane:
BTW, I verified the fix, `lso_*_orphaned_symlink_count` looks better. I will re-test with a final patch.

jsafrane:
I think adding an e2e test could be trivial - just check the metric on the right places.

PR #628, opened 2026-06-03
OCPBUGS-86505: Include in-use volumes in lso_lv_provisioned_PV_count

tsmetana:
The volumes with bind mount were excluded from the lso_lv_provisioned_PV_count metric. This caused the metric to decrease when a volume is used by a pod, which is incorrect. The patch detects valid in-use volumes and adds them back.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate provisioning and reconciliation: rejected devices that correspond to storage classes with already-provisioned volumes are now tracked, improving device status reporting and preventing incorrect provisioning decisions.

* **Tests**
  * Updated tests to reflect the improved reconciliation behavior and ensure metrics are calculated correctly.

### DIFF
--- a/pkg/diskmaker/controllers/lv/reconcile.go
+++ b/pkg/diskmaker/controllers/lv/reconcile.go
@@ -369,8 +369,9 @@ func (r *LocalVolumeReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 		validBlockDevices = append(validBlockDevices, blockDevice)
 	}
 
+	var inUsePVCount map[string]int
 	if len(ignoredDevices) > 0 {
-		r.processRejectedDevicesForDeviceLinks(ctx, ignoredDevices, diskConfig)
+		inUsePVCount = r.processRejectedDevicesForDeviceLinks(ctx, ignoredDevices, diskConfig)
 	}
 
 	r.updateMissingDevicePathMetrics(diskConfig)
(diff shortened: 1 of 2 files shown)


pkg/diskmaker/controllers/lv/reconcile.go
mpatlasov, line 399: We've never dived in this "for-inside-for" body if `len(validBlockDevices) == 0` before this PR. As an optimization, it must be safe to add `if len(validDevices) > 0 {` here leaving localmetrics update outside this `if () {...}`. This way we would update metrics anyway, but save some cycles not executing internal loop if no valid devices detected.

mpatlasov:
@tsmetana , I noticed a weird behavior verifying this PR according to repro steps in JIRA ticket. When pod using local volume is up and running the metric is still equal to 2 -- that's fine. But then, when I deleted pod and pvc, the metric weirdly dropped to 1:
```
$ oc exec -n openshift-local-storage ${DISKMAKER_POD} -c diskmaker-manager -- curl -s http://localhost:8383/metrics | grep lso_lv_provisioned
lso_lv_provisioned_PV_count{nodeName="crc",storageClass="local-test"} 1
```
And in a dozen of seconds returned back to normal:
```
$ oc exec -n openshift-local-storage ${DISKMAKER_POD} -c diskmaker-manager -- curl -s http://localhost:8383/metrics | grep lso_lv_provisioned
lso_lv_provisioned_PV_count{nodeName="crc",storageClass="local-test"} 2
```
This probably means that this PR is not enough to fix the issue. There is something else that triggers such a flip-flop in diskmaker code.

**UPDATE**: Debugging diskmaker during such drop (2-->1), I can see that PV is in "Released" state and diskmaker reports that cleanup is in progress:
```
provisioner_utils.go:162] "PV is still being cleaned, not going to recreate it" pvName="local-pv-123583c" disk="sda"
```
This makes me think that this is **a feature, not a bug**: released PV is not considered as "provisioned", hence it's not counted in `lso_lv_provisioned_PV_count`.

pkg/diskmaker/controllers/lv/reconcile.go
tsmetana: Good point. I think you're correct: running the inner loop in case there are no `validDevices` is just a waste. I'll include your fix and re-test. Thanks!

